PR #12, opened 2021-04-23
Add support for ERC1155

### LINKED ISSUE #9: Add ERC1155

Amxx:
Fixes #9 

This PR add support for the ERC1155 kind, with mintable, pausable, and burnable options.


packages/ui/src/App.svelte
```diff
@@ -73,9 +74,9 @@
         <Tooltip let:trigger text="Coming soon!" placement={overflow ? 'left' : 'bottom'}>
           <button use:trigger class="disabled" aria-disabled="true">ERC777</button>
         </Tooltip>
```
frangio, line 76: Let's remove this one for now? Based on the ongoing discussion about possible deprecation. If we want to keep it it should be at the end of the list now.

frangio (approved):

PR #64, opened 2021-10-07
Add security-contact natspec tag

spalladino:
Adds a new Info section to all contracts in the wizard that includes an optional security-contact input, that renders a security-contact natspect custom tag when set. And since we're at it, the info section also allows the user to tweak the license.

Screenshot from 2021-10-07 16-59-09


packages/ui/src/ERC1155Controls.svelte
```diff
@@ -15,6 +16,7 @@
     mintable: false,
     access: 'ownable',
     upgradeable: false,
+    info: { license: 'MIT '},
```
frangio, line 19: Can you export `infoDefaults` from the core?

packages/ui/src/InfoSection.svelte
```diff
@@ -0,0 +1,32 @@
...
+  <h1>
+    <!-- svelte-ignore a11y-label-has-associated-control -->
+    <label class="flex items-center tooltip-container pr-2">
+      <span>Info</span>
+    </label>
+  </h1>
+
+  <label class="labeled-input">
+    <span class="flex justify-between pr-2">
+      Security email contact
```
frangio, line 19:
```suggestion
      Security Contact
```

packages/ui/src/InfoSection.svelte
```diff
@@ -0,0 +1,32 @@
...
+      <HelpTooltip link="https://github.com/ethereum-lists/contracts/blob/main/README.md">
```
frangio, line 20:
```suggestion
      <HelpTooltip link="https://github.com/ethereum-lists/contracts/blob/main/README.md#tracking-new-deployments">
```

packages/ui/src/InfoSection.svelte
```diff
@@ -0,0 +1,32 @@
...
+      If someone finds a vulnerability in your contract, they can use this email to report it to you. You will need to verify your contract metadata for this to be visible.
```
frangio, line 21: I try to make tooltips as short as possible.

```suggestion
      Where people can contact you to report security issues. Will only be visible if contract metadata is verified.
```

PR #97, opened 2022-03-03
Handle fractional percent quorums

### LINKED ISSUE #93: Wizard bugs out when Quorum is not an Integer in the Governor tab.
See [this](https://forum.openzeppelin.com/t/quorum-fails-if-below-1-in-governor-designer-wizard/25413) OZ Forum post and my reply to it. The values in the source code do not change once a decimal value is added to the input for Quorum %, once it bugs out, the fraction Radio button can also be clicked. See picture attached in the Forum reply. 

A simple solution to this could be to convert the decimal into a fraction - multiply the number by the number of decimal places it has, pass it as the parameter to `GovernorVotesQuorumFraction` while overriding the `quorumDenominator` and multiplying the value by the number of zeroes added to the numerator. This way there would be no need to do input validation on decimals in the form. 

Console Log:
```
print.ts:152 Uncaught (in promise) Error: Number not representable (0.1)
    at tp (print.ts:152:13)
    at Array.map (<anonymous>)
    at print.ts:132:25
    at print.ts:70:21
    at Array.flatMap (<anonymous>)
    at Yu (print.ts:70:8)
    at Xu (print.ts:37:11)
    at Object.e.$$.update (build-generic.ts:31:23)
    at H (index.mjs:1081:12)
    at F (index.mjs:1052:13)
tp @ print.ts:152
(anonymous) @ print.ts:132
(anonymous) @ print.ts:70
Yu @ print.ts:70
Xu @ print.ts:37
e.$$.update @ build-generic.ts:31
H @ index.mjs:1081
F @ index.mjs:1052
Promise.then (async)
Q @ index.mjs:1010
(anonymous) @ index.mjs:1868
(anonymous) @ App.svelte:177
be @ App.svelte:171
(anonymous) @ index.mjs:1866
(anonymous) @ GovernorControls.svelte:109
```

ericglau:
Fixes #93


frangio (approved):
Looks good! Thank you.

PR #102, opened 2022-03-09
Link to contract source from imports

ericglau:
Turns import statements into hyperlinks to the corresponding contract source code on GitHub.

Links to source code rather than docs site for the following reasons:
- Import text should be the hyperlink itself, so only one set of links can be provided (either source or docs but not both)
- Developers using a parent contract may want to see how it was written, so are probably interested in the source
- Each feature's tooltip already has a Read More link that goes to the docs
- If Upgradeability is enabled, the links point to source in [contracts-upgradable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/) so developers can see the initializer functions

### DIFF
--- a/packages/ui/src/App.svelte
+++ b/packages/ui/src/App.svelte
@@ -22,9 +22,9 @@
     import { ContractBuilder, buildGeneric, printContract, printContractVersioned, sanitizeKind, OptionsError } from '@openzeppelin/wizard';
     import { postConfig } from './post-config';
     import { remixURL } from './remix';
-    import { version as contractsVersion } from "@openzeppelin/contracts/package.json";
 
     import { saveAs } from 'file-saver';
+    import { injectHyperlinks } from './utils/inject-hyperlinks';
 
     const dispatch = createEventDispatcher();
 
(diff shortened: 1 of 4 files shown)


packages/ui/src/App.svelte
```diff
@@ -56,7 +57,26 @@
...
+    function injectHyperlinks(code: string) {
+      const importRegex = /(@openzeppelin\/)(contracts-upgradeable\/|contracts\/)(.*)(&quot;)/gm // we are modifying HTML, so use HTML escaped chars
+      let result = code;
+      let match = importRegex.exec(code);
+      while (match != null) {
+        const line = match[0];
+        const ozPrefix = match[1];
+        const contractsLibrary = match[2];
+        const relativePath = match[3];
+        const quotes = match[4];
```
frangio, line 71:
```suggestion
        const [line, ozPrefix, contractsLibrary, relativePath, quotes] = match;
```

packages/ui/src/App.svelte
```diff
@@ -56,7 +57,26 @@
...
+        if (line !== undefined && ozPrefix !== undefined && contractsLibrary !== undefined && relativePath !== undefined && quotes !== undefined) {
```
frangio, line 72: We basically know they won't be undefined, right? Otherwise regex wouldn't have matched. We can use non-null assertions like `${contractsLibrary!}`.

packages/ui/src/App.svelte
```diff
@@ -56,7 +57,26 @@
...
+          const replacedImportLine = `<a href='https://github.com/OpenZeppelin/openzeppelin-${contractsLibrary}blob/v${contractsVersion}/contracts/${relativePath}' target='_blank' rel='noopener noreferrer' style='color: inherit'>${ozPrefix}${contractsLibrary}${relativePath}</a>${quotes}`;
```
frangio, line 73: If we can put the style in a CSS file that would be better.

packages/ui/src/App.svelte
```diff
@@ -56,7 +57,26 @@
     }
 
     $: code = printContract(contract);
-    $: highlightedCode = hljs.highlight('solidity', code).value;
+    $: highlightedCode = injectHyperlinks(hljs.highlight('solidity', code).value);
+
+    function injectHyperlinks(code: string) {
+      const importRegex = /(@openzeppelin\/)(contracts-upgradeable\/|contracts\/)(.*)(&quot;)/gm // we are modifying HTML, so use HTML escaped chars
+      let result = code;
+      let match = importRegex.exec(code);
```
frangio, line 65: Is there a reason why we'd use a while loop instead of just the [replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) method with a `/g` regex?

packages/ui/src/App.svelte
```diff
@@ -56,7 +57,26 @@
     }
 
     $: code = printContract(contract);
-    $: highlightedCode = hljs.highlight('solidity', code).value;
+    $: highlightedCode = injectHyperlinks(hljs.highlight('solidity', code).value);
+
+    function injectHyperlinks(code: string) {
```
frangio, line 62: Let's move this to a separate file under `utils/`.

frangio (changes requested):

packages/ui/src/App.svelte
ericglau: Added CSS for links including hover color.  Any opinion on whether we should keep the underlines?  It adds clutter but makes it obvious that it's a link.

Screen Shot 2022-03-10 at 4 29 02 PM

frangio: I think the underline is fine. The blue on hover was a bit much for me... I removed it in the last commit.

frangio (approved):
LGMT

Issue #101, opened 2022-03-08
Update with support for ERC721Votes


frangio:
- [x] Add "Votes" option in ERC721 tab
- [x] Update "ERC20Votes" option in Governor tab to reflect it also works with ERC721Votes (no changes needed in the code, just the label)

codesport:
@ericglau,

 Under the [ERC721 tab](https://wizard.openzeppelin.com/#erc721), when `Votes `checkbox is selected, I noticed that the votes option still imports `ERC721.sol`.

Would this import be needed?  [draft-ERC721Votes.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.5.0/contracts/token/ERC721/extensions/draft-ERC721Votes.sol) also imports ERC721.sol. 

I'm thinking this is redundant.  But I may be wrong.

ericglau:
Hi @codesport, you are right that since `draft-ERC721Votes.sol` already imports `ERC721.sol`, it isn't strictly necessary to import `ERC721.sol` in the template.  However, it is fine to include it for clarity, since the template inherits from it directly and this can help to avoid errors in case you decide to change the code later to not use the `draft-ERC721Votes.sol` extension.

codesport:
@frangio 

Didn't know imports and inheritance within solidity worked like that without causing headaches.

Definitely appreciate the clarity and your explanation!

PR #103, opened 2022-03-09
Add ERC721Votes support

ericglau:
Part of #101


packages/core/src/erc721.ts
```diff
@@ -126,6 +132,24 @@ function addMintable(c: ContractBuilder, access: Access, incremental = false, ur
...
+  c.addParent(
+    {
+      name: 'EIP712',
+      path: '@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol',
+    },
+    [name, { lit: '"1"' }],
+    true,
+  );
+  c.addOverride('ERC721Votes', functions._afterTokenTransfer);
+  c.addOverride('ERC721Votes', functions._getVotingUnits);
```
frangio, line 150: If I understand correctly we shouldn't need to register this override because it won't conflict with any other module.

packages/core/src/erc721.ts
```diff
@@ -126,6 +132,24 @@ function addMintable(c: ContractBuilder, access: Access, incremental = false, ur
...
+    {
+      name: 'ERC721Votes',
+      path: '@openzeppelin/contracts/token/ERC721/extensions/draft-ERC721Votes.sol',
+    });
+  c.addParent(
+    {
+      name: 'EIP712',
+      path: '@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol',
+    },
+    [name, { lit: '"1"' }],
```
frangio, line 146: Strings are printed with quotes (see [printValue](https://github.com/OpenZeppelin/contracts-wizard/blob/de501648fcc7ea313de179ab5550fff37815eb24/packages/core/src/print.ts#L139)).

```suggestion
    [name, "1"],
```

frangio (changes requested):

frangio (approved):

PR #105, opened 2022-03-18
Install GTM for collecting analytics on actions

shamaleyte:
This page [wizard.openzeppelin.com](http://wizard.openzeppelin.com/) is sending analytics by simple Netlify form submissions. ( See what kind of graphic they are generating in the SS ) .

However, it is a popular website and we are reaching the limits.   

Thus, we need a different mechanism for collecting analytics.
We decided to implement such an analytics process with the help of GTM.

GTM has been installed for collecting analytics on actions.


packages/ui/src/post-config.ts
```diff
@@ -1,13 +1,20 @@
...
+  }
+}
 const instance = uuid();
 
 export type Action = 'copy' | 'remix' | 'download-npm' | 'download-vendored';
 
 // NOTE: We have to make sure any fields sent in the body are defined in the
 // hidden form in public/index.html.
 export async function postConfig(opts: Required<GenericOptions>, action: Action) {
+  console.log("running postConfig... ");
```
frangio, line 16: Is console.logging these messages necessary? I would remove them just to not clutter the console.

packages/ui/src/post-config.ts
```diff
@@ -21,7 +28,19 @@ export async function postConfig(opts: Required<GenericOptions>, action: Action)
     }),
   });
 }
-
+function sendAnalyticsToGTM(opts: any, action : Action){
```
frangio, line 31: Have you been able to confirm this works? I believe the GTM script tags should be inserted not in index.html but in embed.html which is the context in which `postConfig` will execute. Note that the application really runs in embed.html and index.html contains an iframe with the application.

shamaleyte: @frangio  just update it as suggested. 

Please check and approve?

frangio (approved):

ericglau:
@shamaleyte @frangio Should this also be added to https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/ui/public/cairo.html (which is accessed from https://wizard.openzeppelin.com/cairo)?

frangio:
IMO it should be enough to just add this in embed.html, but I wanted to move this PR forward and reevaluate that later. If embed.html is enough then there should be no need to add to Cairo.

PR #106, opened 2022-03-21
Separate Governor options for ERC20Votes and ERC721Votes

ericglau:
- Separate Governor options for ERC20Votes and ERC721Votes
- Governor's ERC721Votes option disables the `Token decimals` setting and sets it to 0.  This will cause the `Proposal Threshold` value and absolute `Quorum` values to be used as is (without the exponent from token decimals).
- Allow user to specify 0 token decimals (even for ERC20Votes or COMP-like).


packages/ui/src/GovernorControls.svelte
```diff
@@ -112,13 +112,15 @@
     {/if}
   </label>
 
+  {#if opts.votes !== 'erc721votes'}
```
frangio, line 115: An `if` is a little too much, it causes the form elements to move around. We should add the `disabled` attribute to the `input` element and set it to 0 when ERC721 is used. It can be a little difficult to implement that last part in Svelte, you may need to do something like we do for mintable and incremental in the ERC721 component. Or there may be a simpler way to do it.

frangio (changes requested):

packages/ui/src/GovernorControls.svelte
```diff
@@ -115,9 +137,9 @@
   <p class="tooltip-container flex justify-between items-center pr-2">
     <label class="text-sm">
       Token decimals:
-      <input type="number" bind:value={opts.decimals} placeholder={defaults.decimals.toString()} class="input-inline" use:resizeToFit>
+      <input {...decimalsInputAttributes} type="number" bind:value={opts.decimals} placeholder={defaults.decimals.toString()} class="input-inline" use:resizeToFit use:error={errors?.decimals}>
```
frangio, line 140: We should have a variable `disabledDecimals: boolean`, here we can then say `disabled={disabledDecimals}`.

frangio (approved):

PR #112, opened 2022-04-08
Add a Custom tab

### LINKED ISSUE #82: Add global toggle for "Access Control" section
... like the Upgradeability section, and make sure that if it is enabled then the contract is actually inherited. Currently even if Ownable is selected, unless a feature like Mintable makes use of it the resulting contract will not have Ownable and this could be confusing for users.

### LINKED ISSUE #110: Add a "General" tab
We want to add a tab that is more of a "blank canvas" boilerplate builder so people can choose independent features and build an Ownable + UUPSUpgradeable contract easily that is not necessarily an ERC20 or any of the other options.

Depends on https://github.com/OpenZeppelin/contracts-wizard/issues/82 for the access control section to make sense.

### LINKED ISSUE #133: No option for ERC1155 without owner
There is a `setURI` `onlyOwner` (or role) function by default that can't be removed. We don't have an option for owner-less ERC1155.

frangio:
Fixes #110
Fixes #82
Fixes #133 

Adds a Custom contract type ("Custom" UI tab and `custom` contract type for API):
- Pausable, Access Control, and Upgradeability sections.

Adds toggle for Access Control:
- Defaults to `false`.
- If certain features are selected that require access control, makes Access Control section required (checked and greyed out button) and defaults to `ownable`.
- Adds API function `isAccessControlRequired` for every contract type --> used for UI, and can be used by API consumers to determine in what cases would access control be required.

Adds toggle for ERC1155 `setURI`:
- Adds "Updatable URI" option, `true` by default (which implies Access Control is enabled by default for ERC1155, but can be disabled by deselecting Updatable URI)


ericglau:
@frangio Please review (I can't add you as reviewer because you opened this PR)

Particularly I am wondering what your thought is for the `isAccessControlRequired` function in the API -- currently I've made the function available for all contract types (including ERC1155 and Governor).  

However, access control is always needed in ERC1155 (due to `setURI`) and is never needed for Governor (because it doesn't use it).  
- Should we remove the function for these cases (or at least remove it for Governor)?
- Should ERC1155 have `access` option of type `'ownable'|'roles'` (e.g. exclude `false` from the type?)

ericglau:
Ready for review

packages/core/src/erc20.ts
```diff
@@ -55,6 +55,10 @@ export function printERC20(opts: ERC20Options = defaults): string {
   return printContract(buildERC20(opts));
 }
 
+export function isAccessControlRequired(opts: Partial<ERC20Options>): boolean {
+  return opts.mintable || opts.pausable || opts.snapshots || opts.upgradeable === 'uups';
```
frangio, line 59: What mechanism do we have to make sure we keep this list up to date?

Can we extend `core/src/test.ts` so it also checks the consistency of this option with the results somehow? For every combination. Is that overkill?

packages/core/src/generate/erc1155.ts
```diff
@@ -13,6 +13,7 @@ const blueprint = {
   pausable: booleans,
   mintable: booleans,
   supply: booleans,
+  updatableUri: booleans,
```
frangio, line 16: I think we need to add a `generate/general.ts`.

packages/core/src/print.ts
```diff
@@ -87,22 +87,27 @@ function printConstructor(contract: Contract, helpers: Helpers): Lines[] {
...
     }
-  } else {
+  } else if (!helpers.upgradeable) {
     return [];
+  } else {
+    return INITIALIZE_IMPLEMENTATION;
   }
 }
 
+const INITIALIZE_IMPLEMENTATION =
```
frangio, line 101: This is more like "DISABLE_INITIALIZERS" or something. "INITIALIZE_IMPLEMENTATION" doesn't seem accurate.

packages/ui/src/inputs/ToggleRadio.svelte
```diff
@@ -1,25 +1,26 @@
...
     }
 
-    wasEnabled = enabled;
+    wasChecked = checked;
     wasValue = value || wasValue;
   }
 </script>
 
-<input type="checkbox" bind:checked={enabled}>
+<input type="checkbox" bind:checked={checked} disabled={disabled}>
```
frangio, line 26: Shorthand syntax:

```suggestion
<input type="checkbox" bind:checked {disabled}>
```

packages/ui/src/AccessControlSection.svelte
```diff
@@ -1,13 +1,32 @@
...
 
+  import ToggleRadio from './inputs/ToggleRadio.svelte';
   import HelpTooltip from './HelpTooltip.svelte';
 
   export let access: Access;
+  export let requireAccessControl: boolean;
+
+  $: {
+    if (access === false && requireAccessControl) {
+      access = 'ownable';
```
frangio, line 12: It would be nicer UX if we could set it to the previously selected value instead of always Ownable. It's a little tricky to implement but ToggleRadio implements kind of similar behavior so it can probably work as inspiration.

packages/ui/src/ERC1155Controls.svelte
```diff
@@ -13,6 +13,11 @@
     ...erc1155.defaults,
     info: { ...infoDefaults }, // create new object since Info is nested
   };
+
+  let requireAccessControl: boolean;
+  $: {
+    requireAccessControl = erc1155.isAccessControlRequired(opts);
+  }
```
frangio, line 20: Shorthand:

```suggestion
  $: requireAccessControl = erc1155.isAccessControlRequired(opts);
```

[Docs](https://svelte.dev/docs#component-format-script-3-$-marks-a-statement-as-reactive)

packages/ui/src/ERC1155Controls.svelte
```diff
@@ -64,10 +69,17 @@
         Useful for emergency response.
       </HelpTooltip>
     </label>
+    <label class:checked={opts.updatableUri}>
+      <input type="checkbox" bind:checked={opts.updatableUri}>
+      Updatable URI
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts/4.x/api/token/erc1155#ERC1155-_setURI-string-">
+        Privileged accounts will be able to set a new URI for all token types. Clients will replace any instance of &lbrace;id&rbrace; in the URI with the tokenId.
```
frangio, line 76: May be nicer to read? In case you didn't consider it.

```suggestion
        Privileged accounts will be able to set a new URI for all token types. Clients will replace any instance of {"{id}"} in the URI with the tokenId.
```

packages/ui/src/ERC1155Controls.svelte
```diff
@@ -64,10 +69,17 @@
...
+      </HelpTooltip>
+    </label>
   </div>
 </section>
 
-<AccessControlSection bind:access={opts.access} />
+<AccessControlSection bind:access={opts.access} requireAccessControl={requireAccessControl} />
```
frangio, line 82:
```suggestion
<AccessControlSection bind:access={opts.access} {requireAccessControl} />
```

packages/ui/src/ERC20Controls.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
   import HelpTooltip from './HelpTooltip.svelte';
 
-  import type { KindedOptions } from '@openzeppelin/wizard';
+  import type { KindedOptions, } from '@openzeppelin/wizard';
```
frangio, line 4: :thinking:

ericglau (approved):

PR #123, opened 2022-05-03
Add Cairo to GitHub actions

ericglau:
Add steps to compile and test `cairo-core`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,12 @@ on:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        package:
+          - core
+          - core-cairo
+
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -16,12 +16,18 @@ jobs:
...
+        run: yarn tsc
+        working-directory: packages/core-cairo
       - name: Check Svelte
         run: yarn svelte-check
         working-directory: packages/ui
-      - name: Run tests
+      - name: Run tests - Solidity
         run: yarn test
         working-directory: packages/core
+      - name: Run tests - Cairo
```
frangio, line 31: I think it would be better to run tests in separate jobs, so that they show up as separate checks for Cairo and Solidity (I imagine we will rarely update both at the same time and it will be more useful to quickly check that changes to Solidity didn't affect Cairo, and so on). We need to duplicate some of the setup steps which is fine.

frangio (changes requested):

frangio (approved):
Amazing!

Issue #115, opened 2022-04-12
Add initializer call for ERC20Votes


ericglau:
When selecting ERC20 with Votes and Upgradeability, the initializer should call `__ERC20Votes_init()` as a best practice.

PR #124, opened 2022-05-03
Add initializer call for ERC20Votes

ericglau:
- Add initializer call for ERC20Votes
- Fix SafetyCheck.sol to use upgradeable contracts

Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/115


frangio (approved):
Good catch on using the upgradeable dependency.

PR #127, opened 2022-05-11
Wizard API for Cairo

ericglau:


packages/ui/src/cairo/ERC20Controls.svelte
```diff
@@ -1,26 +1,17 @@
...
-    symbol: 'MTK',
-    burnable: false,
-    pausable: false,
-    premint: '0',
-    decimals: '18',
-    mintable: false,
-    access: 'ownable',
-    upgradeable: false,
-    info: { ...infoDefaults },
+    ...erc20defaults,
```
frangio, line 13: There's a potential problem with the `info` value. `erc20defaults.info` is a nested object that will not be copied, so it will result in side effects when we mutate it by writing to `opts.info[someKey]`. We should make a deep clone of `erc20defaults`. There is no built-in function to do this except `structuredClone` which is a new standard so we would need a [polyfill](https://github.com/ungap/structured-clone)), or alternatively we can use `const deepClone = obj => JSON.parse(JSON.stringify(obj))` which is hacky but common.

frangio (approved):
LGTM!

Issue #129, opened 2022-05-13
Wizard API for Solidity


ericglau:
Solidity equivalent of https://github.com/OpenZeppelin/contracts-wizard/pull/127

rg12301:
**Hello ser!**
This comment is to request you for an earlier release of **Wizard API for Solidity**. My team is participating in a hackathon for which we are building a one-click solution to deploy smart contracts. We found no better option other than OpenZeppelin Contracts as they are secure and well audited. If my request could not be fulfilled in a week or two, please guide me with a workaround for the same.
**Also, accept my gratitude for your awesome work.**

ericglau:
@rg12301 Thanks for the input. We'll aim to have something available for you to use within the next couple of weeks.

rg12301:
Awesome!! thank you for considering my request.

ericglau:
@rg12301 The API would look like [this](https://github.com/ericglau/contracts-wizard/blob/solidityapi/packages/core/README.md), let me know if you have any feedback on it.

You could build the project from [this source](https://github.com/ericglau/contracts-wizard/tree/solidityapi) if you would like to try it out in the meantime before it is released.

rg12301:
Cool! I'll try it and let you know.

rg12301:
@ericglau I was able to build and print smart contracts using the API. However, I got stuck when tried to deploy the application to Vercel. Deployment failed due to the following error:

![image](https://user-images.githubusercontent.com/56386618/173533296-52b01658-1b38-4a6e-b8a4-ddc49facb330.png)

As in the above ss, this issue is in the `packages/core/src/contract.ts:92:16`. I also tried after modifying the `tsconfig.json` as described, but the error continues to appear.

ericglau:
@rg12301 The package has been published to npm: https://www.npmjs.com/package/@openzeppelin/wizard

You can try installing the package with `npm install @openzeppelin/wizard` and see if that resolves your issue.

rg12301:
Superb!! it is working now. You were just in time, only two days for final submission.
Hearty thanks 😄🙏🏼

PR #136, opened 2022-06-07
Wizard API for Solidity, update API format for Cairo

ericglau:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/129

Uses API format like `erc20.print()` and `erc20.defaults` for both Solidity and Cairo.


packages/core-cairo/src/api.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  defaults: CommonOptions;
+}
+
+export interface ERC20 extends WizardContractAPI {
+  defaults: Required<ERC20Options>;
+}
+export interface ERC721 extends WizardContractAPI {
+  defaults: Required<ERC721Options>;
+}
```
frangio, line 22: I believe this encodes what we want better:

```suggestion
export interface WizardContractAPI<Options extends CommonOptions> {
  /**
   * Returns a string representation of a contract generated using the provided options. If opts is not provided, uses `defaults`.
   */
  print: (opts?: Options) => string,
  
  /**
   * The default options that are used for `print`.
   */
  defaults: Required<Options>;
}

export type ERC20 = WizardContractAPI<ERC20Options>;
export type ERC721 = WizardContractAPI<ERC721Options>;
```

packages/core/src/api.ts
```diff
@@ -0,0 +1,47 @@
+import type { CommonOptions } from './common-options';
+import { printERC20, defaults as erc20defaults, ERC20Options } from './erc20';
+import { printERC721, defaults as erc721defaults, ERC721Options } from './erc721';
+import { printERC1155, defaults as erc1155defaults, ERC1155Options } from './erc1155';
+import { printGovernor, defaults as governorDefaults, GovernorOptions } from './governor';
+
+export interface WizardContractAPI {
```
frangio, line 7: Same type as for Cairo.

frangio (approved):
LGTM. Thanks!

PR #137, opened 2022-06-07
Add getInitialSupply API for Cairo

ericglau:
After using the Wizard API to generate a contract, users may want to generate test cases to ensure their contracts behave (and continue to behave) as expected.  

This may include checking the initial supply of ERC20 tokens.  Provide a helper function to calculate the initial supply based on the user-specified premint amount and number of decimals.

Needed for https://github.com/onlydustxyz/generator-starknet/issues/30

Depends on https://github.com/OpenZeppelin/contracts-wizard/pull/136 for new API format.


ericglau:
Closing for now since this needs to be built on top of #136

PR #142, opened 2022-06-23
Cairo - Add UUPS radio button

ericglau:
Under Upgradeability section, add a UUPS radio button as the only option.

This affects the UI only.  The API for `upgradeable` option will remain as `false | true` to remain backward compatible.

### DIFF
--- a/packages/ui/src/cairo/UpgradeabilitySection.svelte
+++ b/packages/ui/src/cairo/UpgradeabilitySection.svelte
@@ -2,18 +2,12 @@
   import type { Upgradeable } from '@openzeppelin/wizard-cairo';
 
   import HelpTooltip from '../HelpTooltip.svelte';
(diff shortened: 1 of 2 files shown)


frangio:
Have you seen the `ToggleRadio` element we're using for Solidity? I think that's what we should be using here no?

https://github.com/OpenZeppelin/contracts-wizard/blob/cd15091a9f3dacde38a2a36c3c7d53a76d896416/packages/ui/src/UpgradeabilitySection.svelte#L16

ericglau:
Changed it to use ToggleRadio.  It would still need a separate variable for the UI (`upgradeableWithKind`) since ToggleRadio works with strings (unless we make an alternative ToggleRadio that takes `true`).

frangio (approved):

Issue #141, opened 2022-06-20
Cairo - add Roles-based access control


ericglau:
Add roles-based access control, pending https://github.com/OpenZeppelin/cairo-contracts/issues/249

PR #147, opened 2022-07-05
Cairo - Add Role-Based Access Control

ericglau:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/141


packages/core-cairo/src/contract.ts
```diff
@@ -35,6 +36,7 @@ export interface BaseFunction {
   name: string;
   implicitArgs?: Argument[];
   args: Argument[];
+  literalArgs?: string[];
```
frangio, line 39: I don't think this is used here. Should be removed.

packages/core-cairo/src/contract.ts
```diff
@@ -44,11 +46,16 @@ export interface BaseFunction {
...
 export interface ContractFunction extends BaseFunction {
-  libraryCalls: BaseFunction[];
+  libraryCalls: LibraryCall[];
   code: string[];
   final: boolean;
 }
 
+export interface LibraryCall {
+  baseFn: BaseFunction;
+  literalArgs: string[];
```
frangio, line 56: We could be using the `Value` type here.

Why `literalArgs` instead of just `args`?

packages/core-cairo/src/print.ts
```diff
@@ -184,9 +185,11 @@ function printFunction(fn: ContractFunction): Lines[] {
 
   const returnArgs = fn.returns?.map(a => typeof a === 'string' ? a : a.name);
 
-  fn.libraryCalls.forEach(callFn => {
-    const libraryCall = `${getFunctionName(callFn)}()`;
-    code.push(libraryCall);
+  fn.libraryCalls.forEach(libraryCall => {
+    const libraryCallString = libraryCall.literalArgs ? 
+      `${getFunctionName(libraryCall.baseFn)}(${libraryCall.literalArgs.join(', ')})`:
+      `${getFunctionName(libraryCall.baseFn)}()`;
```
frangio, line 191: `literalArgs` is always an array as far as I can tell.

```suggestion
    const libraryCallString = 
      `${getFunctionName(libraryCall.baseFn)}(${libraryCall.literalArgs.join(', ')})`;
```

packages/core-cairo/src/set-access-control.ts
```diff
@@ -74,4 +125,63 @@ const functions = defineFunctions({
     args: [],
     parentFunctionName: 'renounce_ownership',
   },
+
+  hasRole: {
+    module: modules.AccessControl,
+    kind: 'view' as const,
```
frangio, line 131: `as const` isn't required here. `defineFunctions` expects its argument to have a specific type, so it will infer `kind` to be the right type.

packages/core-cairo/src/set-access-control.ts
```diff
@@ -38,14 +55,43 @@ export function requireAccessControl(c: ContractBuilder, fn: BaseFunction, acces
       c.addLibraryCall(functions.assert_only_owner, fn);
       break;
     }
+    case 'roles': {
+      const roleId = role + '_ROLE';
+      if (c.addVariable(`const ${roleId} = ${to251BitHash(roleId)}`)) {
```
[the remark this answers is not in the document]
ericglau: I don't see a way to avoid using a sentence.
Changed it to say "truncated to 251 bits" -- do we need to be clear that it's the first 251 bits? (I think truncated somewhat implies that)

packages/core-cairo/src/contract.ts
ericglau: `Value` type could be useful if we have more use cases for this interface later on, but I don't see a benefit in using that at this point.

Changed to `args`.

packages/core-cairo/src/set-access-control.ts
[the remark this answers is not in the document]
ericglau: Changed to `# keccak256('PAUSER_ROLE')[0:251 bits]`

frangio (approved):
LGTM!

PR #148, opened 2022-07-12
Add npm badges

ericglau:
Add npm badges for wizard and wizard-cairo

Colors are based on css for the titles [here](https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/ui/public/standalone.css#L64) and [here](https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/ui/public/standalone.css#L79)


frangio (approved):
Nice!

PR #156, opened 2022-07-21
Fix access checkbox when implied by default

### LINKED ISSUE #152: ERC1155 access control default in UI does not look enabled
ERC1155 has Updatable URI enabled by default, which implies (forces) Access Control to be enabled.

In the UI, Access Control is greyed out but does not appear enabled (even though it is enabled in the resulting code).
Screen Shot 2022-07-18 at 10 11 05 AM

ericglau:
Fixes #152 

If `access` should be implied (e.g. `required` is true), then set it to be enabled regardless of whether it was previously implied (e.g. regardless of `wasRequired`)

### DIFF
--- a/packages/ui/src/AccessControlSection.svelte
+++ b/packages/ui/src/AccessControlSection.svelte
@@ -14,13 +14,11 @@
   $: {
     if (wasRequired && !required) {
       access = wasAccess;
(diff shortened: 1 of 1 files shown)


frangio (approved):
Nice work!

PR #170, opened 2022-08-16
Cairo - Add Ownable `owner` view

### LINKED ISSUE #165: Cairo - Ownable missing owner view
Ownable should have the following function:
```
@view
func owner{
        syscall_ptr : felt*,
        pedersen_ptr : HashBuiltin*,
        range_check_ptr
    }() -> (owner: felt):
    let (owner: felt) = Ownable.owner()
    return (owner)
end
```

ericglau:
Fixes #165


packages/core-cairo/src/erc721.test.ts.md
```diff
@@ -1066,6 +1086,16 @@ Generated by [AVA](https://avajs.dev).
...
     end␊
     ␊
+    @view␊
+    func owner{␊
+            syscall_ptr: felt*,␊
+            pedersen_ptr: HashBuiltin*,␊
+            range_check_ptr␊
+        }() -> (owner: felt):␊
+        let (owner) = Ownable.owner()␊
+        return (owner)␊
```
frangio, line 1096: Can you remind me if it was not possible to just do `return Ownable.owner()` ?

frangio (approved):

packages/core-cairo/src/erc721.test.ts.md
ericglau: `return Ownable.owner()` works in this case because `Ownable.owner()` also returns `owner`.

But if the library's return value was named differently, then this doesn't work.  I think it's because this would be considered the [tail recursion scenario](https://www.cairo-lang.org/docs/reference/syntax.html#call-statement).

Issue #176, opened 2022-09-05
Add support for solidity ERC777 token


satya1807:
Add ERC777 token with
- access control [ownable, roles]
- mintable
- upgradeable
- premint

frangio:
We are not interested in support ERC-777 in Contracts Wizard at the moment. There is a proposal to deprecate the contract and we might go forward with that route. See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2620

PR #179, opened 2022-09-06
Solidity ERC777 contract

jveer634:
Added erc777 token with options:
- mintable
- premint
- access
- pausable
- upgradeable

NOTE: ERC777 Contains burnable functionality by default.


jveer634:
closes #176

frangio:
We are not interested in support ERC-777 in Contracts Wizard at the moment. There is a proposal to deprecate the contract and we might go forward with that route. See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2620

Issue #187, opened 2022-10-24
Download Hardhat development environment 


ericglau:
Provide an option to download a Hardhat development environment based on the configured contract, so that the user can build it with just a few commands and start developing on it further.

ericglau:
[Design document](https://docs.google.com/document/d/1K_UGlnJRIPJ7JbN0J9p8hyg5mhY_AS93WDl2usV5auQ)

PR #188, opened 2022-10-25
Download Hardhat package

ericglau:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/187


packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+
+const config: HardhatUserConfig = {
+  solidity: {
+    version: "0.8.4",
+    settings: {
+      optimizer: {
+        enabled: true,
+      },
+    },
+  }
```
frangio, line 20: :eyes: 
```suggestion
  },
```

packages/core/src/environments/hardhat/upgradeable/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "hardhat-sample",
+  "version": "0.0.1",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "npx hardhat test"
```
frangio, line 7: `npx` not necessary here. When you say `npm run foo`, the script `foo` will run with $PATH pointing to the dependency executables.

```suggestion
    "test": "hardhat test"
```

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
+import JSZip from "jszip";
+import type { GenericOptions } from "./build-generic";
+import type { Contract } from "./contract";
+import type { ERC1155Options } from "./erc1155";
+import { printContract } from "./print";
+
+const hardhatConfig = (upgradeable: boolean) => `\
+import { HardhatUserConfig } from "hardhat/config";
+import "@nomicfoundation/hardhat-toolbox";${upgradeable ? `
+import "@openzeppelin/hardhat-upgrades";` : ''}
```
frangio, line 10: I prefer the extra newline. :slightly_smiling_face: We can also use `formatLines` to simplify some of this (see comment on this later).

```suggestion
import "@nomicfoundation/hardhat-toolbox";
${upgradeable ? `import "@openzeppelin/hardhat-upgrades";` : ''}
```

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+  "compilerOptions": {
+    "target": "es2020",
+    "module": "commonjs",
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "skipLibCheck": true
+  }
+}
+`
```
frangio, line 37: There are missing semicolons throughout this file.

```suggestion
`;
```

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+typechain-types
+
+#Hardhat files
+cache
+artifacts
+`
+
+const test = (c: Contract, opts?: GenericOptions) => {
+
+let buf = `\
```
frangio, line 54: What's with the lack of indentation in this function?

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+import { expect } from "chai";
+import { ethers${c.upgradeable ? ', upgrades' : ''} } from "hardhat";
```
frangio, line 56: This or something like this. Also applies in `script`.

```suggestion
let plugins = ['ethers'];
if (c.upgadeable) {
  plugins.push('upgrades');
}

let buf = `\
import { expect } from "chai";
import { ${plugins.join(', ')} } from "hardhat";
```

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+
+describe("${c.name}", function () {
+  it("Test contract", async function () {
+    const ContractFactory = await ethers.getContractFactory("${c.name}");
+
+    const instance = await ${c.upgradeable ? 'upgrades.deployProxy(ContractFactory);' : 'ContractFactory.deploy();' }
```
frangio, line 62: Also applies in `script`.

```suggestion
    const instance = await ${c.upgradeable ? 'upgrades.deployProxy(ContractFactory)' : 'ContractFactory.deploy()' };
```

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+
+if (opts !== undefined) {
+  switch (opts.kind) {
+    case 'ERC20':
+    case 'ERC721':
+      buf += `    expect(await instance.name()).to.equal("${opts.name}");`
+      break;
+
+    case 'ERC1155':
+      buf += `    expect(await instance.uri(0)).to.equal("${(opts as ERC1155Options).uri}");`
```
frangio, line 75: Cast not necessary, it gets narrowed automatically (I checked).

```suggestion
      buf += `    expect(await instance.uri(0)).to.equal("${opts.uri}");`
```

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+    const instance = await ${c.upgradeable ? 'upgrades.deployProxy(ContractFactory);' : 'ContractFactory.deploy();' }
+    await instance.deployed();
+
+`;
+
+if (opts !== undefined) {
+  switch (opts.kind) {
+    case 'ERC20':
+    case 'ERC721':
+      buf += `    expect(await instance.name()).to.equal("${opts.name}");`
```
frangio, line 71: I'm not opposed to keeping it this way (string concatenation), but `formatLines` allows us to build long strings with indentation in a nicer way, IMO. Not sure if you considered using it.

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+`
+
+const readme = `\
+# Sample Hardhat Project
+
+This project demonstrates a basic Hardhat use case. It comes with a contract generated by [OpenZeppelin Wizard](https://wizard.openzeppelin.com/), a test for that contract, and a script that deploys that contract.
+
+## Installing dependencies
+
+\`\`\`shell
```
frangio, line 124: I tried recently and adding `shell` here doesn't do anything really. We can remove it.

packages/core/src/environments/hardhat/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "hardhat-sample",
+  "version": "0.0.1",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "npx hardhat test"
+  },
+  "author": "",
+  "license": "ISC",
```
frangio, line 10: I believe we should replace this dynamically with the license the user puts in the Wizard.

packages/core/src/environments/hardhat/package-lock.json
```diff
@@ -0,0 +1,6369 @@
+{
+  "name": "hardhat-sample",
+  "version": "0.0.1",
+  "lockfileVersion": 1,
```
frangio, line 4: What version of npm are you using? This is an old lockfile format. We should see `lockfileVersion: 2`. You should be using npm 8.

packages/core/src/zipHardhat.ts
```diff
@@ -0,0 +1,160 @@
...
+import { printContract } from "./print";
+
+const hardhatConfig = (upgradeable: boolean) => `\
+import { HardhatUserConfig } from "hardhat/config";
+import "@nomicfoundation/hardhat-toolbox";${upgradeable ? `
+import "@openzeppelin/hardhat-upgrades";` : ''}
+
+const config: HardhatUserConfig = {
+  solidity: {
+    version: "0.8.4",
```
frangio, line 14: We should unify this with both of these:

https://github.com/OpenZeppelin/contracts-wizard/blob/f0ee5fc80ccd497559a5a5be1de29ffd487bca4d/packages/core/hardhat.config.js#L46
https://github.com/OpenZeppelin/contracts-wizard/blob/f0ee5fc80ccd497559a5a5be1de29ffd487bca4d/packages/core/src/print.ts#L9

We can define the constant in a dedicated file and use it in all three places.

frangio (commented):
Very good. Minor comments.

packages/ui/src/App.svelte
```diff
@@ -102,6 +102,18 @@
...
       }
     };
+
+    const zipEnvModule = import('@openzeppelin/wizard/zipEnv');
+
+    const downloadHardhatHandler = async () => {
+      const { zipHardhat } = await zipEnvModule;
+      const zip = await zipHardhat(contract, opts);
+      const blob = await zip.generateAsync({ type: 'blob' });
+      saveAs(blob, 'hardhat.zip');
```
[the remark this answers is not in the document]
ericglau: Since the vendored zip is `contracts.zip`, I think `project.zip` would align with that.

packages/core/zipEnv.ts
```diff
@@ -0,0 +1 @@
+export * from './src/zipHardhat';
```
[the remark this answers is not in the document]
ericglau: Renamed to `zip-env` and `zip-hardhat`

Could not get dynamic imports to work if the extension is named as `.d.ts`.  We should look at that separately.

packages/core/src/zipHardhat.ts
ericglau: Used formatLines for `test` but kept everything else as strings since those are simple enough as-is.

ericglau:
Addressed review comments, also added snapshots to the tests

ericglau:
Re: socket-security, scripts and native code look fine as part of `classic-level`, and bin scripts for different contexts.

packages/core/package.json
```diff
@@ -16,7 +16,8 @@
     "prepublish": "rimraf dist *.tsbuildinfo && hardhat clean",
     "test": "ava",
     "test:watch": "ava --watch",
-    "version": "node ../../scripts/bump-changelog.js"
+    "version": "node ../../scripts/bump-changelog.js",
+    "update-env": "npm install --prefix ./src/environments/hardhat && npm install --prefix ./src/environments/hardhat/upgradeable"
```
frangio, line 20: Is this supposed to update the lockfiles? AFAICT it won't.

packages/core/src/zip-hardhat.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    'test/',
+    'test/test.ts',
+    'tsconfig.json',
+  ]);
+}
+
+async function extractAndRunPackage(zip: JSZip, t: ExecutionContext<unknown>) {
+  const files = Object.values(zip.files);
+
+  let tempFolder = await fs.mkdtemp(path.join(os.tmpdir(), 'openzeppelin-wizard-'));
```
frangio, line 81:
```suggestion
  const tempFolder = await fs.mkdtemp(path.join(os.tmpdir(), 'openzeppelin-wizard-'));
```

packages/core/src/zip-hardhat.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  t.snapshot(contentComparison);
+}
+
+async function getItemString(zip: JSZip, key: string) {
+  const obj = zip.files[key];
+  if (obj !== undefined) {
+    return await asString(obj);
+  } else {
+    return '';
```
frangio, line 118: Any reason not to throw an error here?

packages/core/src/zip-hardhat.test.ts.md
```diff
@@ -0,0 +1,521 @@
...
+      pragma solidity ^0.8.4;␊
+      ␊
+      import "@openzeppelin/contracts/token/ERC20/ERC20.sol";␊
+      ␊
+      contract MyToken is ERC20 {␊
+          constructor() ERC20("My Token", "MTK") {}␊
+      }␊
+      `,
+      `import { HardhatUserConfig } from "hardhat/config";␊
+      import "@nomicfoundation/hardhat-toolbox";␊
```
frangio, line 22: We seem to be missing Hardhat Chai Matchers in the config.

packages/core/src/zip-hardhat.ts
```diff
@@ -47,60 +48,77 @@ typechain-types
...
+    2,
+    ...spaceBetween(
+      getImports(c),
+      getTestCase(c),
+    ),
+  );
+
+  function getTestCase(c: Contract) {
+    let testcase = [];
+    testcase.push(
```
frangio, line 64: There is just one thing being pushed, can we return the array directly?

frangio (commented):
Final comments

packages/core/src/zip-hardhat.test.ts.md
ericglau: `hardhat-toolbox` includes `hardhat-chai-matchers`

frangio (approved):

PR #209, opened 2022-12-20
Update links to Contracts for Cairo v0.5.1

ericglau:


frangio:
If this is only for docs links it may be better to use the unversioned links, which redirect to latest. What do you think?

frangio (approved):

ericglau:
Linking to a specific version helps to avoid confusion if there are unreleased changes in the libraries, which may differ in usage from what is available in Wizard.

PR #269, opened 2023-07-05
Use ethers v6

ericglau:
Use ethers v6, hardhat-toolbox v3, and latest versions of Hardhat and `@openzeppelin/hardhat-upgrades` when downloading Hardhat development package.


packages/core/src/zip-hardhat.ts
```diff
@@ -32,7 +32,8 @@ const tsConfig = `\
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "strict": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "resolveJsonModule": true
```
frangio, line 36: What causes this to be necessary?

ericglau: This was added to sample projects generated by Hardhat in `https://github.com/NomicFoundation/hardhat/pull/3659`

frangio (approved):

PR #271, opened 2023-07-10
Add option to download Foundry project

ericglau:


packages/core/src/zip-foundry.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const command = `cd "${tempFolder}" && ${setGitUser} && ${setup} && ${test} && ${script}`;
+
+    const exec = util.promisify(child.exec);
+    const result = await exec(command);
+
+    t.regex(result.stdout, /1 passed/);
+    t.regex(result.stdout, /Deploying contract to /);
+  } finally {
+    await rimraf(tempFolder);
```
frangio, line 104: So you know, it is also possible to use [`t.teardown`](https://github.com/avajs/ava/blob/main/docs/02-execution-context.md#tteardownfn). I don't think there is any difference though.

packages/core/src/zip-foundry.ts
```diff
@@ -0,0 +1,288 @@
...
+        id: build
+
+      - name: Run Forge tests
+        run: |
+          forge test -vvv
+        id: test
+`;
+
+const dotEnv = `\
+PRIVATE_KEY=`
```
frangio, line 47: What is this for? It's usually bad practice to put the private key in an env file so I don't know that we should have this.

packages/core/src/zip-foundry.test.ts
frangio, line 1: We based this on the defaults from `forge init` today. How can we make sure that we continue to provide the standard initial default? Or at least that it doesn't change too much?

packages/ui/src/App.svelte
```diff
@@ -108,17 +108,29 @@
...
-    const zipEnvModule = import('@openzeppelin/wizard/zip-env');
+    const zipHardhatModule = import('@openzeppelin/wizard/zip-env-hardhat');
 
     const downloadHardhatHandler = async () => {
-      const { zipHardhat } = await zipEnvModule;
+      const { zipHardhat } = await zipHardhatModule;
       const zip = await zipHardhat(contract, opts);
       const blob = await zip.generateAsync({ type: 'blob' });
-      saveAs(blob, 'project.zip');
+      saveAs(blob, 'hardhat.zip');
```
frangio, line 117: I would name both `project.zip`, or something Wizard-specific. This makes the zip file look like it contains Hardhat/Foundry.

packages/ui/src/App.svelte
```diff
@@ -108,17 +108,29 @@
...
         await postConfig(opts, 'download-hardhat', language);
       }
     };
+
+    const zipFoundryModule = import('@openzeppelin/wizard/zip-env-foundry');
+
+    const downloadFoundryHandler = async () => {
+      const { zipFoundry } = await zipFoundryModule;
+      const zip = await zipFoundry(contract, opts);
+      const blob = await zip.generateAsync({ type: 'blob', platform: 'UNIX' });
```
frangio, line 128: Why is `platform: 'UNIX'` needed here?

packages/ui/src/App.svelte
```diff
@@ -200,7 +212,17 @@
           <ZipIcon />
           <div class="download-option-content">
             <p>Development Package (Hardhat)</p>
-            <p>Sample project to get started with development and testing.</p>
+            <p>Sample Hardhat project to get started with development and testing.</p>
+          </div>
+        </button>
+        {/if}
+
+        {#if opts?.kind !== "Governor" && opts?.upgradeable === false}
```
frangio, line 220: What was the reason why Governor is disallowed?

packages/core/zip-env-foundry.js
frangio, line 1: There is a better way to do this now with [package entry points](https://nodejs.org/api/packages.html#package-entry-points). Something like the following in the package.json:

```json
"exports": {
  "zip-env-foundry": "./dist/zip-foundry.js"
}
```

Types should be resolved correctly too without need of the `.ts` file.

I would try this separately though because the `"exports"` object needs to be larger and I'm not sure what else needs to be in it, or what other changes may need to be done (TS Config?).

packages/core/src/zip-foundry.test.ts
ericglau: `forge init` generates contents based on its own [templates](https://github.com/foundry-rs/foundry/tree/master/cli/assets).  The only relevant files from there are the workflow template and gitignore template.  (And foundry.toml which is also generated but not stored as a template).

Some options:
1. Perform `forge init` within setup.sh and then delete the contract/test/script files that we don't need.  This would also allow us to remove the git repo setup logic from setup.sh, since `forge init` covers that.  The drawback is that the package that the user downloads does not have everything (it is missing some files until the user runs setup.sh), which is potentially confusing.
2. Use a git submodule in the Wizard repo itself to track the foundry repo.  I think this would be a bit excessive.
3. Watch for changes in the foundry repo in some automated way.
4. Omit workflow template (gitignore seems useful though).

I think we can do (1).  What do you think?

packages/ui/src/App.svelte
ericglau: Governor requires a token contract to be deployed as well, and that token contract is not defined anywhere.  Therefore  any provided script or test would not work until the user provides a token contract.

Should we do this anyways and provide instructions in the downloaded package (e.g. tell user to also download an ERC20/721 Votes contract as "Single file")?

packages/core/src/zip-foundry.test.ts
ericglau: Changed it to use `beforeEach` and `afterEach` instead.

packages/core/zip-env-foundry.js
ericglau: Opened separate issue in https://github.com/OpenZeppelin/contracts-wizard/issues/273

packages/core/src/zip-foundry.ts
ericglau: Removed, so now the script can only simulate a deployment locally (which is enough for the sample project). Users can read the provided scripting link if they want to learn more.

packages/ui/src/App.svelte
frangio: No strong opinion on this. I think it's fine to disable it for now, it would also be fine to enable it with a warning saying that they should also download the token.

packages/core/src/zip-foundry.test.ts
frangio: Makes sense to me. What would be included in the zip file if we run forge init in setup.sh ?

ericglau: It would have only the contract, script, test, README.md, and setup.sh.

packages/ui/src/App.svelte
ericglau: Leaving it disabled for now.

packages/core/src/zip-foundry.test.ts
frangio: Yeah I think that makes sense.

packages/core/src/zip-foundry.ts
```diff
@@ -0,0 +1,223 @@
...
+          '}',
+        ],
+      ),
+      '}',
+    ];
+  }
+};
+
+const setupSh = `\
+# Check if git is installed
```
frangio, line 134:
```suggestion
#!/usr/bin/env bash

# Check if git is installed
```

packages/core/src/zip-foundry.ts
```diff
@@ -0,0 +1,223 @@
...
+This project demonstrates a basic Foundry use case. It comes with a contract generated by [OpenZeppelin Wizard](https://wizard.openzeppelin.com/), a test for that contract, and a script that deploys that contract.
+
+## Installing Foundry
+
+See [Foundry installation guide](https://book.getfoundry.sh/getting-started/installation).
+
+## Initializing the project
+
+\`\`\`
+./setup.sh
```
frangio, line 193:
```suggestion
bash setup.sh
```

This works too and we wouldn't need to make the file executable. It feels somewhat cleaner to me.

packages/core/src/zip-foundry.ts
```diff
@@ -0,0 +1,223 @@
...
+        ],
+      ),
+      '}',
+    ];
+  }
+};
+
+const setupSh = `\
+# Check if git is installed
+if ! [ -x "$(command -v git)" ]
```
frangio, line 135:
```suggestion
if ! command -v git &> /dev/null
```

packages/core/src/zip-foundry.ts
```diff
@@ -0,0 +1,223 @@
...
+then
+  echo "git command not found. Install git and try again."
+  exit 1
+fi
+
+# Check if Foundry is installed
+if ! [ -x "$(command -v forge)" ]
```
frangio, line 142:
```suggestion
if ! command -v forge &> /dev/null
```

packages/core/src/zip-foundry.ts
```diff
@@ -0,0 +1,223 @@
...
+  # Add remappings
+  if [ -f "remappings.txt" ]
+  then
+    echo "" >> remappings.txt
+  fi
+  echo "@openzeppelin/=lib/openzeppelin-contracts/" >> remappings.txt
+
+  # Perform initial git commit
+  git add .
+  git commit -m "Initial commit" --quiet
```
frangio, line 173: `forge init` already generates initial commits. This kind of message seems more in line with those commits:

```suggestion
  git commit -m "openzeppelin: add wizard output" --quiet
```

I don't like it fully but I didn't know what else to put before the colon?

frangio (approved):
Minor comments.

packages/core/src/zip-foundry.ts
ericglau: This didn't work on the Github Action when I tried it.  See https://github.com/OpenZeppelin/contracts-wizard/actions/runs/5536781265/jobs/10104887999 -- it somehow went into the if-statement to print `git command not found. Install git and try again.` but also printed the git path `/usr/bin/git`

frangio: It may be because `command` is a builtin, not sure though.

We could instead use `which` (it's what I would have used initially).

```suggestion
if ! which git &> /dev/null
```

packages/core/src/zip-foundry.ts
ericglau: Currently we are using `forge init` with `--no-commit`.

The challenge is Wizard output is already in the directory before we run `forge init`.  So the initial commit would include the Wizard output anyways, regardless of when we do the commit.

We also need to remove template contracts, and add remappings.txt.  I consider it neatest to commit everything at the same time.

(The alternative would be something like this, but it feels excessive:
commit 1: forge init along with Wizard output (let forge init do the commit)
commit 2: remove templates
commit 3: install OZ contracts and add remappings)

ericglau: I agree with changing the commit message format to be more similar to what `forge` normally does though.

frangio: Ah, I wasn't thinking of all that. It makes sense to use `--no-commit` and generate a single commit. I would consider perhaps adding some mention of wizard in the initial commit message, that's all.

packages/core/src/zip-foundry.ts
ericglau: Changed to use `which`

packages/core/src/zip-foundry.ts
ericglau: Changed to use `which`

frangio (approved):

Issue #281, opened 2023-08-23
Enable ERC20Permit by default


SKYBITDev3:
`ERC20Permit` should exist in all newly-designed ERC20 tokens to save gas.

frangio:
What do you mean "to save gas"?

SKYBITDev3:
Im the process of a token holder changing allowance, `approve` was the usual way to do it, but it requires a transaction, which costs gas. `permit` also allows changing allowance, but only requires a signature, which costs no gas, so it's a better way. New tokens should have `permit` as a more cost-effective way to approve allowance than `approve`.

Dapps these days that work with tokens check whether `permit` exists in the token, e.g. by looking for selector `d505accf`: https://github.com/Uniswap/uniswapx-sdk/blob/e555ce2ffdf98028ae6ca869354c28f29f602b6f/abis/MockERC20.json#L409

Only if permit doesn't exist in a particular token then they fall back to calling `approve`.

frangio:
"Costs no gas" is not really accurate, but I see that the Wizard tooltip is claiming something similar. It should be cheaper though.

I would definitely recommend including ERC20Permit in a token, but for UX reasons. I agree with making it a default in Wizard.

SKYBITDev3:
> "Costs no gas" is not really accurate

Why do you say it's not really accurate? I've used signatures as a user of platforms like OpenSea and Rarible, as well as developed dapps that check for `permit` in the token and calls it instead of calling `approve`. On the UI instead of a transaction screen, some info about the signature request is shown (particularly the amount) and a "sign" button. After pressing it the allowance approval is done, then a screen appears for the transfer transaction (which of course costs gas).

frangio:
The signature doesn't cost anything. The transfer function will be slightly more expensive when it includes a permit. Overall it will probably be cheaper than having a separate approve transactoin.

SKYBITDev3:
I've just tried `permit` on Avalanche Fuji and found that balance actually decreased slightly (less than 1c worth), and trying it with 0 balance fails with "insufficient funds" error. So you are right that cost isn't exactly 0.

SKYBITDev3:
The important point to note is that *anyone* can call `permit`, and it's the caller that pays the gas.
If the token owner calls `permit`, it costs him gas.
If the spender calls `permit`, it costs her gas, so then the approval process is free for the owner, which makes it much better for user experience.

Here's some of my code in which spender calls `permit`:

```js
[...]

PR #284, opened 2023-09-16
Update OpenZeppelin Contracts to 5.0

frangio:
Leaving this as Draft until Contracts 5.0 is released.

Preview: https://deploy-preview-284--openzeppelin-contracts-wizard.netlify.app/

- Uses OpenZeppelin Contracts 5.0
- Requires constructor or initializer arguments for initial owner or role assignments if using access control.
- Removes Download Vendored ZIP option in favour of Hardhat or Foundry package options.
- Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/285
- Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/281 

To do when 5.0 is released:
- [x] Update package dependencies (remove rc) from `package.json` and `packages/core/src/environments/hardhat` 
- [x] Update doc links (remove rc)
- [x] From each of these folders, run `npm i --package-lock-only`, then commit the updated `package-lock.json` files
  - [x] `packages/core/src/environments/hardhat`
  - [x] `packages/core/src/environments/hardhat/upgradeable`
- [x] Run `yarn ava --update-snapshots` and commit
- [ ] Merge PR
- [ ] Release Wizard API on npm and github


packages/core/src/contract.ts
```diff
@@ -100,10 +103,10 @@ export class ContractBuilder implements Contract {
     });
   }
 
-  get imports(): string[] {
+  get imports(): ParentContract[] {
     return [
-      ...[...this.parentMap.values()].map(p => p.contract.path),
-      ...this.using.map(u => u.library.path),
+      ...[...this.parentMap.values()].map(p => p.contract),
```
Amxx, line 108: I'm was wondering why we don't `...this.parentMap.values().map(p => p.contract)`, and then I realised `parentMap` is a `Map<string, Parent>` and not a `Record<string, Parent>`. Any reason why we use Map here ?

packages/core/src/erc1155.test.ts.md
```diff
@@ -30,19 +31,19 @@ Generated by [AVA](https://avajs.dev).
...
         bytes32 public constant URI_SETTER_ROLE = keccak256("URI_SETTER_ROLE");␊
     ␊
-        constructor()␊
+        constructor(address defaultAdmin, address uriSetter)␊
             ERC1155("https://gateway.pinata.cloud/ipfs/QmcP9hxrnC1T5ATPmq2saFeAM1ypFX9BnAswCdHB9JCjLA/")␊
         {␊
-            _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);␊
-            _grantRole(URI_SETTER_ROLE, msg.sender);␊
+            _grantRole(DEFAULT_ADMIN_ROLE, defaultAdmin);␊
+            _grantRole(URI_SETTER_ROLE, uriSetter);␊
```
Amxx, line 46: we could simplify the constructor by removing the `uriSetter` and not grant the `URI_SETTER_ROLE` role at construction. The `defaultAdmin` would have to grant the role before anyone can set the URI.

packages/core/src/erc20.test.ts.md
```diff
@@ -97,14 +40,19 @@ Generated by [AVA](https://avajs.dev).
...
 
     `// SPDX-License-Identifier: MIT␊
-    pragma solidity ^0.8.9;␊
+    pragma solidity ^0.8.20;␊
     ␊
     import "@openzeppelin/contracts/token/ERC20/ERC20.sol";␊
-    import "@openzeppelin/contracts/security/Pausable.sol";␊
+    import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Pausable.sol";␊
     import "@openzeppelin/contracts/access/Ownable.sol";␊
+    import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Permit.sol";␊
```
Amxx, line 48: Not an issue, but ordering is strange. Ideally we would have all the `token/ERC20/**` and then the access control

packages/core/src/erc721.ts
```diff
@@ -1,6 +1,6 @@
-import { Contract, ContractBuilder } from './contract';
+import { Contract, ContractBuilder, ParentContract } from './contract';
```
Amxx, line 1: I don't see this new import being used

Amxx (approved):
Left some "design" remarks, that are not really actionnable.
PR LGTM.

packages/core/src/contract.ts
ericglau: While we could use Record, I think a benefit of Map is that it maintains insertion order which helps ensure consistency of the code output.

packages/core/src/erc20.test.ts.md
ericglau: This happens because pause functions need access control, so when ERC20Pausable is added, we add Ownable before adding the other ERC20 features.

This affects imports, inheritance order, and parent initializer calls.  If we want to sort, we should sort all of them for consistency.

packages/core/src/contract.ts
frangio: > I think a benefit of Map is that it maintains insertion order which helps ensure consistency of the code output

Yeah I think this is a reasonable argument. Object/Record is not completely useless if ordering is relevant, but it's unreliable in some edge cases: https://stackoverflow.com/a/5525820

packages/core/src/erc1155.ts
```diff
@@ -88,55 +88,67 @@ export function buildERC1155(opts: ERC1155Options): Contract {
...
-    },
-    [uri],
-  );
-
-  c.addOverride('ERC1155', functions._beforeTokenTransfer);
-  c.addOverride('ERC1155', supportsInterface);
+  const ERC1155 = {
+    name: 'ERC1155',
+    path: '@openzeppelin/contracts/token/ERC1155/ERC1155.sol',
+    transpiled: true,
```
frangio, line 94: It would make sense to set a default for `transpiled` in `addParent`, so we don't have to write it out always. The default could be: if it looks like an interface (`/^I[A-Z]/`) set `transpiled: false`, otherwise `transpiled: true`.

packages/ui/src/AccessControlSection.svelte
```diff
@@ -36,7 +36,7 @@
       <span class="ml-1">
         <ToggleRadio bind:value={access} defaultValue="ownable" disabled={required} />
       </span>
-      <HelpTooltip align="right" link="https://docs.openzeppelin.com/contracts/4.x/api/access">
+      <HelpTooltip align="right" link="https://docs.openzeppelin.com/contracts/5.x-rc/api/access">
```
frangio, line 39: This link will redirect to the latest version:

```suggestion
      <HelpTooltip align="right" link="https://docs.openzeppelin.com/contracts/api/access">
```

frangio:
Looks good on cursory review.

packages/core/src/infer-transpiled.ts
```diff
@@ -0,0 +1,5 @@
+import type { ReferencedContract } from "./contract";
+
+export function inferTranspiled(c: ReferencedContract): boolean {
+  return c.transpiled !== undefined ? c.transpiled : c.name.match(/^I[A-Z]/) === null;
```
frangio, line 4:
```suggestion
  return c.transpiled ?? !/^I[A-Z]/.test(c.name)
```

stebansaa:
This may be generating an infinite gas error on the remix ide.

ericglau:
@stebansaa Please to open a separate issue with details of your steps and the error that you are getting. Thanks.

PR #298, opened 2023-10-16
Add AccessManaged

ericglau:


Amxx (approved):
LGTM

Amxx:
Just wondering how the constructor argument will show up in the UI.

Things like name, symbol, premint are hardcoded, and they don't show up in the constructor argument. What is the logic for what should/shouldn't be a constructor argument ?

ericglau:
The constructor argument would need to be passed through, e.g.
```
contract MyToken is ERC20, AccessManaged {
    constructor(address initialAuthority)
        ERC20("MyToken", "MTK")
        AccessManaged(initialAuthority)
    {}
}
```

We specifically add this as a constructor argument to the Wizard-generated contract using [`addConstructorArgument`](https://github.com/OpenZeppelin/contracts-wizard/pull/298/files#diff-fad73bd75d660270ecd84f280c0d31d6abadb222a2f74e7d7b042674adca70dfR35)

ernestognw (approved):
LGTM, thanks @ericglau

PR #306, opened 2023-11-29
Wiz Merge

zackrw:
Merge for upcoming release


.gitignore
```diff
@@ -1,3 +1,10 @@
 dist
 *.tsbuildinfo
 node_modules
+
+# Local Netlify folder
+.netlify
+.vscode
```
ericglau, line 7: gitignores for environment specific files are typically not added to a repo, but in a global gitignore that you define locally.  Does anything in this repo specifically create a .netlify folder?
```suggestion
```

.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "deno.enable": true,
+  "deno.enablePaths": [
+    "packages/ui/api"
+  ],
+  "deno.unstable": true,
+  "deno.importMap": ".netlify/edge-functions-import-map.json",
+  "deno.path": "/Users/zack/Library/Preferences/netlify/deno-cli/deno"
+}
```
ericglau, line 9: I think we can remove this file

package-lock.json
```diff
@@ -0,0 +1,12148 @@
+{
```
ericglau, line 1: We use yarn for this repo rather than npm. We should remove this package-lock.json file to avoid conflicts.

packages/core/package.json
```diff
@@ -33,6 +33,7 @@
     "typescript": "^4.9.0"
   },
   "dependencies": {
-    "array.prototype.flatmap": "^1.2.4"
+    "array.prototype.flatmap": "^1.2.4",
+    "openai": "^4.6.0"
```
ericglau, line 37: The openai dependency doesn't seem used in core
```suggestion
    "array.prototype.flatmap": "^1.2.4"
```

package.json
```diff
@@ -15,5 +15,8 @@
   },
   "devDependencies": {
     "wsrun": "^5.2.4"
+  },
+  "dependencies": {
+    "axios": "^1.5.0"
```
ericglau, line 20: This dependency doesn't seem to be used
```suggestion
```

yarn.lock
```diff
@@ -4,25 +4,25 @@
 
 "@alloc/quick-lru@^5.2.0":
   version "5.2.0"
-  resolved "https://registry.yarnpkg.com/@alloc/quick-lru/-/quick-lru-5.2.0.tgz#7bf68b20c0a350f936915fcae06f58e32007ce30"
+  resolved "https://registry.npmjs.org/@alloc/quick-lru/-/quick-lru-5.2.0.tgz"
```
ericglau, line 7: Was there a need to change the domains, or can you regenerate this file using yarn?

ericglau (commented):
Some initial comments

.vscode/settings.json
zackrw: Removed.

package-lock.json
zackrw: Removed

yarn.lock
zackrw: Regenerated with `yarn`

packages/ui/api/ai.ts
```diff
@@ -0,0 +1,79 @@
+import OpenAI from 'https://esm.sh/openai@4.11.0'
+import { OpenAIStream, StreamingTextResponse } from 'https://esm.sh/ai@2.2.16'
+import { erc20Function, erc721Function, erc1155Function, governorFunction, customFunction } from '../src/wiz-functions.ts'
+import { Redis } from 'https://esm.sh/@upstash/redis@1.25.1'
```
ericglau, line 4: Are these imports using URLs because they are part of an edge function? If so, are these dependencies still needed in `packages/ui/package.json`?

zackrw: Good point. I think we can remove from package.json then.

packages/ui/src/Wiz.svelte
```diff
@@ -0,0 +1,246 @@
...
+        const chunk = decoder.decode(value);
+        result += chunk
+
+        if (result.startsWith('{"function_call":')) {
+          currentMessage = 'Executing function...'
+        }
+        else {
+          currentMessage = result
+        }
+      }
```
ericglau, line 94: Since reader is not being changed, it would be clearer to have an if-statement and then a nested while-loop:
```suggestion
      if (reader) {
        while (true) {
          const { done, value } = await reader.read();
          if (done) {
            break;
          }
          // Massage and parse the chunk of data
          const chunk = decoder.decode(value);
          result += chunk
  
          if (result.startsWith('{"function_call":')) {
            currentMessage = 'Executing function...'
          }
          else {
            currentMessage = result
          }
        }
      }
```

packages/ui/src/Wiz.svelte
```diff
@@ -0,0 +1,246 @@
...
+      {#each messages as message}
+        <div class="flex gap-2">
+          <div class="relative h-[36px] w-[36px] flex-none">
+            <div class="absolute w-[36px] h-[36px] glimmery rounded-full animate-spin-slow"></div>
+            {#if message.role === 'assistant'}
+              <div class="absolute m-[2px] flex items-center justify-center w-[32px] h-[32px] bg-gray-900 text-gray-50 rounded-full">
+                <WizAvatar />
+              </div>
+            {/if}
+            {#if message.role === 'user'}
```
ericglau, line 192:
```suggestion
            {:else if message.role === 'user'}
```

packages/ui/src/Wiz.svelte
```diff
@@ -0,0 +1,246 @@
...
+            {/if}
+          </div>
+          {#if message.role === 'assistant'}
+            {#if message.content}
+              <div class="text-sm bg-gray-200 rounded-md p-4">{message.content}</div>
+            {:else}
+              <div class="text-sm bg-gray-200 rounded-md font-bold p-4 text-gray-500">{'Called a function'}</div>
+            {/if}
+          {/if}
+          {#if message.role === 'user'}
```
ericglau, line 205:
```suggestion
          {:else if message.role === 'user'}
```

packages/ui/src/Wiz.svelte
zackrw: Applied

packages/ui/src/Wiz.svelte
zackrw: Applied

packages/ui/src/Wiz.svelte
zackrw: Applied

ericglau (approved):
LGTM!

PR #324, opened 2024-02-12
Cairo: Remove non-standard increase_allowance and decrease_allowance from ERC20

ericglau:
Remove safe allowance option and functions as per https://github.com/OpenZeppelin/cairo-contracts/pull/881


andrew-fleming (commented):
Looking good, Eric! I think there's a few other remnants of `SafeAllowance` to get rid of:

[1, 2](https://github.com/ericglau/contracts-wizard/blob/removeSafeAllowance/packages/core-cairo/src/erc20.ts#L360-L407)
[3](https://github.com/ericglau/contracts-wizard/blob/removeSafeAllowance/packages/core-cairo/src/generate/erc20.ts#L15)

ericglau:
@andrew-fleming Thanks for noticing! Updated.

andrew-fleming:
Ah another thing: it's now deprecated to add the `external` attribute on generated trait impls:

```
    #[generate_trait]
    #[external(v0)]
    impl ExternalImpl of ExternalTrait {
        foo() {}
    }
```

Instead of using the `external` attribute, we can now use `abi(per_item)` and add the `external` attr to each function within the trait e.g.

```
    #[generate_trait]
    #[abi(per_item)]
    impl ExternalImpl of ExternalTrait {
        #[external(v0)]
        fn foo() {}
    }
```

Here's an [example](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/tests/mocks/account_mocks.cairo#L135-L148) in the lib

The alternative is to use loose functions and not define them within a trait. Having the external trait doesn't really matter AFAICT. We need to make a decision on this for the wizard @martriay @ericnordelo

ericglau:
@andrew-fleming I made the change suggested in your comment above.

Pending further decision on whether we should use loose functions instead, we could address that in a separate issue if needed.

PR #325, opened 2024-02-13
Temporarily disable Wizard AI

ericglau:
There is a configuration issue with Wizard AI, so temporarily disabling the AI popup until it is fixed.


ernestognw (approved):
Thanks for the prompt response @ericglau !

Issue #333, opened 2024-02-21
Consider making upgradeability opt out


martriay:
We should consider this possibility based following on https://github.com/OpenZeppelin/cairo-contracts/issues/921

ericglau:
I agree, if that is our recommendation to users, we should make it enabled by default in Wizard.

martriay:
Bear in mind this hasn't been developed yet on the Cairo side, we should hold this until we do.

PR #334, opened 2024-02-22
Enable upgradeable option by default for Cairo

ericglau:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/333


ericglau:
The socket-security check does not look right.  There were no dependency changes for this PR.

ericnordelo (approved):
LGTM!

PR #335, opened 2024-02-23
Show error page if embed version out of supported range

ericglau:
In our documentation sites, we embed Wizard for different versions of the Contracts libraries, for example on https://docs.openzeppelin.com/contracts-cairo/0.9.0/wizard

However, Wizard is only compatible with specific versions of the libraries, which is specified (and shown on Wizard) as a semantic version string.

In the embeddings, we should allow passing through a `version` parameter for the selected version of the documentation.  If this version is outside of the compatible versions, the embed will show an error page instead of the actual Wizard.  By doing this, we will no longer need to manually remove Wizard from older versions of the doc sites.

This version can be passed in from a documentation page like this:
```
<oz-wizard style="display: block; min-height: 40rem;" data-lang="cairo" version="0.8.0"></oz-wizard>
```

Example result:
https://deploy-preview-335--openzeppelin-contracts-wizard.netlify.app/embed?lang=cairo&version=0.8.0

### DIFF
--- a/packages/core-cairo/src/index.ts
+++ b/packages/core-cairo/src/index.ts
@@ -20,6 +20,6 @@ export { OptionsError } from './error';
 export type { Kind } from './kind';
 export { sanitizeKind } from './kind';
 
-export { contractsVersion, contractsVersionTag } from './utils/version';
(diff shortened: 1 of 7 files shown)


ericnordelo (approved):
Looking great Eric!

PR #337, opened 2024-02-26
Remove Cairo lang version

ericglau:
Output code comments currently include compatible Contracts version and Cairo lang version.

However, different versions of Contracts may require different versions of Cairo lang.

We should only display the Contracts version, and users should determine the version of Cairo lang to use based on that.  For example, this can be found from the `cairo-version` field in [Scarb.toml](https://github.com/OpenZeppelin/cairo-contracts/blob/main/Scarb.toml).


andrew-fleming (approved):
LGTM!

PR #346, opened 2024-03-15
Add download option for upgradeable Foundry package

ericglau:
Adds option to download Foundry package when upgradeability is enabled.
Uses [Foundry Upgrades](https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades) to deploy a UUPS or transparent proxy, depending on what proxy kind is selected.


packages/core/src/zip-foundry.ts
```diff
@@ -232,15 +338,15 @@ bash setup.sh
 ## Testing the contract
 
 \`\`\`
-forge test
+forge test${c.upgradeable ? ' --ffi --force' : ''}
```
ernestognw, line 341: What about adding `ffi` to the `foundry.toml` configuration instead of hardcoding the flag?

https://book.getfoundry.sh/reference/config/testing?highlight=ffi#ffi

packages/core/src/zip-foundry.ts
```diff
@@ -186,23 +268,47 @@ if ! [ -f "foundry.toml" ]
...
 
   # Add remappings
   if [ -f "remappings.txt" ]
   then
     echo "" >> remappings.txt
   fi
-  echo "@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/" >> remappings.txt
+${c.upgradeable ? `\
+  echo "@openzeppelin/contracts/=lib/openzeppelin-contracts-upgradeable/lib/openzeppelin-contracts/contracts/" >> remappings.txt
+  echo "@openzeppelin/contracts-upgradeable/=lib/openzeppelin-contracts-upgradeable/contracts/" >> remappings.txt
```
ernestognw, line 302: Shall we add a remapping for foundry upgrades as well?

ericglau: That shouldn't be needed as Foundry Upgrades is intended to work with the default mappings.

packages/core/src/zip-foundry.ts
ericglau: I like this idea. Moved it to foundry.toml.

ernestognw (approved):
All good, thanks @ericglau

Issue #225, opened 2023-02-21
Consider adding Governor and Votes clock options


frangio:
OpenZeppelin Contracts 4.9 will support governors and voting tokens based on timestamps as opposed to block numbers.

ericglau:
This should include:
- Adding an option under Votes for ERC20 and ERC721
- Adding an option in Governor, so that GovernorSettings is initialized using timestamp-compatible values.

PR #347, opened 2024-03-18
Add timestamp based Governor and Votes clock options

ericglau:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/225

- [x] For ERC20 and ERC721, under the Votes setting (which is now its own section), adds Block Number (default) and Timestamp options.
- [x] For Governor, under a new Token Clock Mode section, adds Block Number (default) and Timestamp options.  Users must choose this to match what they are using in their token.  If Timestamp is selected, GovernorSettings is initialized using timestamp-compatible values.

~~- [ ] Link the above options together if possible.~~ I don't think we should link these options at this time. Current Wizard behaviour does not link settings between tabs, and users may not realize if certain settings are linked. For example, users may not notice if Wizard modifies their ERC20 votes option when they are on the Governor tab.


packages/core/src/utils/duration.ts
```diff
@@ -32,3 +32,20 @@ export function durationToBlocks(duration: string, blockTime: number): number {
...
+  const match = duration.trim().match(durationPattern);
+
+  if (!match) {
+    throw new Error('Bad duration format');
+  }
+
+  const value = match[1]!;
+  const unit = match[2]! as DurationUnit;
+
+  if (unit === 'block' || unit === 'month' || unit === 'year') {
```
Amxx, line 46: what about hour / day ?

Amxx (approved):
LGTM

PR #355, opened 2024-04-23
Cairo: Use Contracts for Cairo v0.12.0, add ERC20Votes

ericglau:
- Use Contracts for Cairo v0.12.0
- Use Hooks with ERC20
- Add Votes option to ERC20
- Sort implemented traits for consistency


packages/core-cairo/src/erc20.ts
```diff
@@ -214,6 +223,78 @@ function addMintable(c: ContractBuilder, access: Access) {
...
+    tags: [],
+  };
+  c.addImplementedTrait(SNIP12Metadata);
+
+  c.addFunction(SNIP12Metadata, {
+    name: 'name',
+    args: [],
+    returns: 'felt252',
+    code: [
+      "'DAPP_NAME'"
```
ericnordelo, line 246: Users should be able to set this and DAPP_VERSION as inputs.

packages/core-cairo/src/erc20.ts
```diff
@@ -214,6 +223,78 @@ function addMintable(c: ContractBuilder, access: Access) {
   requireAccessControl(c, externalTrait, functions.mint, access, 'MINTER', 'minter');
 }
 
+function addVotes(c: ContractBuilder) {
```
ericnordelo, line 226: The order of the embedded implementations changes when the Mintable option is toggled. Not a big deal, but it would be nice to maintain the consistency when using the same components.

ericnordelo (commented):
Looking good! I Left a couple of comments.

packages/core-cairo/src/erc20.ts
ericglau: Changed it to use the contract name as the name, and added a separate version input.  

Let me know if you think there is a need to allow a different name (i.e. a separate name input specifically for SNIP12) than the contract name.

packages/core-cairo/src/erc20.ts
ericglau: Can you share an example of which embedded impls you are referring to?

andrew-fleming (commented):
Now that we have hooks, I'm thinking that we can embed `ERC20MixinImpl` with pausable enabled and add `assert_not_paused` in the `before_update` hook.

This will change the behavior from past Cairo wizard iterations because approvals will be able to go through when paused. Do note that the extension in Solidity also allows approvals when paused. As an added bonus if we agree on the change, the code will be much cleaner. What do you guys think? @ericglau @ericnordelo

ericglau:
@andrew-fleming That sounds good to me! However, I would prefer for that to be in a separate PR, and perhaps (depending on timing) it may make sense to do this when hooks are available for the other contract kinds. I think the current PR is an improvement over the current state (and enables 0.12.0) and therefore makes sense on its own.

andrew-fleming:
> I would prefer for that to be in a separate PR, and perhaps (depending on timing) it may make sense to do this when hooks are available for the other contract kinds. I think the current PR is an improvement over the current state (and enables 0.12.0) and therefore makes sense on its own.

Sounds good on all accounts!

ericglau:
@ericnordelo @andrew-fleming I've added SNIP12-specific options as "Application Name" and "Application Version", and added checks to ensure these are not empty string.   This PR is ready for re-review.

ericnordelo (approved):
Looking good Eric! One small detail, I think we could move SNIP12MetadataImpl above the ERC20HooksTraitImpl just for visibility.

PR #359, opened 2024-05-21
Support Contracts for Cairo v0.13.0, use hooks with ERC721/1155 and Pausable

ericglau:
- Use OpenZeppelin Contracts for Cairo v0.13.0.
- Use Hooks with ERC721 and ERC1155.
- Use Hooks for Pausable.


ericnordelo (approved):
LGTM!

PR #367, opened 2024-06-18
Remove uuid dependency from UI

ericglau:
This dependency does not seem to be used, so it can be removed.

Replaces / Closes https://github.com/OpenZeppelin/contracts-wizard/pull/365


frangio (approved):
Makes sense. We used it for the previous analytics implementation that was removed a while back.

PR #368, opened 2024-06-19
Add Cairo compilation workflow

ericnordelo:
This PR adds a script for updating a local scarb "test_project" required for compiling the files since we don't have a javascript compiler yet, and we need scarb for module resolution (importing the oz library).

This PR also adds a workflow that generates the scarb project and tries to compile it.

How to test this:

1. From project root, yarn install
2. Install scarb using asdf (the version for the current PR is 2.6.5)
3. packages/core-cairo directory, run `yarn run update_scarb_project`
4. From packages/core-cairo directory, run `scarb build`

Note that some contracts are failing to compile (missing a ContractAddress import). I plan to add a fix for those on a separate PR that will also migrate to the library version 0.14.0.


.github/workflows/cairo-compilation.yml
```diff
@@ -0,0 +1,34 @@
+name: Compile Cairo contracts
```
ericglau, line 1: Note that there is already a `.github/workflows/test.yml` workflow which sets up prerequisites and runs all tests, which includes all languages in its test matrix (Solidity and Cairo).  The tests would include `packages/core-cairo/src/test.ts` to generate all possible contracts, and we may want to run Cairo compilations as part of that.

Should we prefer combining the workflow and compilation steps into the existing test as described above?  Or do we see a benefit in having a separate workflow and not performing Cairo compilations during the regular test runs?

packages/core-cairo/src/generate/sources.ts
```diff
@@ -85,17 +85,26 @@ function generateContractSubset(subset: Subset, kind?: Kind): GeneratedContract[
   }
 }
 
-export function* generateSources(subset: Subset, kind?: Kind): Generator<GeneratedSource> {
+export function* generateSources(subset: Subset, uniqueName?: boolean, kind?: Kind): Generator<GeneratedSource> {
+  let counter = 1;
   for (const c of generateContractSubset(subset, kind)) {
+    if (uniqueName) {
```
ericglau, line 91: It seems there are two different concepts here: (1) the file name (*.cairo), and (2) the contract name (used in the contract source code)

1. The file name normally uses the `id` field of the `GeneratedContract` interface.  This is generated based on a hash of the selected options, so supposedly it should be unique for each different combination of options.

2. The contract name is normally `MyToken` or `MyContract` as defined in the `blueprint` for each kind, e.g. in `packages/core-cairo/src/generate/erc20.ts`.

Do the contract names in (2) need to be unique?  And even if so, could we still use the original `id` hash as the file names?

packages/core-cairo/src/test.ts
```diff
@@ -33,7 +33,7 @@ async function testCompile(t: ExecutionContext<Context>, kind: keyof KindedOptio
   // TODO define output path for compilation environment
   const generatedSourcesPath = path.join('artifacts/contracts', `generated`);
   await fs.rm(generatedSourcesPath, { force: true, recursive: true });
-  await writeGeneratedSources(generatedSourcesPath, 'all', kind);
+  await writeGeneratedSources(generatedSourcesPath, 'all', false, kind);
```
ericglau, line 36: Why is `uniqueName` set as `false` here, but `true` in `update-scarb-project.ts`?

packages/core-cairo/test_project/Scarb.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "test_project"
+version = "0.1.0"
+edition = "2023_01"
+cairo-version = "2.6.4"
+scarb-version = "2.6.5"
+
+[dependencies]
+starknet = "2.6.4"
+openzeppelin = { git = "https://github.com/OpenZeppelin/cairo-contracts.git", tag = "v0.13.0" }
```
ericglau, line 10: The Contracts for Cairo version is also stored in the `contractsVersionTag` constant in `packages/core-cairo/src/utils/version.ts`.

It would be good if we could have only one place to define that version, and reference that wherever needed.  I'm not sure how feasible that is though.

ericglau (commented):
Thanks, this is looking good! Just left some comments/questions.

.github/workflows/cairo-compilation.yml
ericnordelo: The issue I see is that we don't have a Cairo compiler in javascript, so we would need to call scarb as a bash script or similar from javascript to keep the same flow we follow in Solidity, and I don't think that's better than just having a different workflow directly using the scarb cli instead.

packages/core-cairo/src/generate/sources.ts
ericnordelo: The filename is also the module name that needs to be imported in the lib.cairo file, and there are some restrictions like not having numbers at the beginning, which current Ids have.

The contract names need to be unique for Scarb to compile them together. Note that the proposed solution is building a scarb project and compiling it since we currently don't have a friendly API for compiling from javascript (or even from rust).

packages/core-cairo/src/test.ts
ericnordelo: I kept the flow you had in the tests the same and this one wasn't using the uniqueNames, but the ids. The unique name is a requirement for compiling, and that's why it is passed as true in update-scarb-project.ts.

packages/core-cairo/src/generate/sources.ts
ericnordelo: Do you think could be helpful to add the Id to the uniqueName like `Contract12_123231313123`?

packages/core-cairo/test_project/Scarb.toml
ericnordelo: Maybe we can update this file as part of the update-scarb-project script to set the version from that file as well.

packages/core-cairo/src/test.ts
ericglau: As in my other comment, we could keep both of these consistent and always use unique names.

packages/core-cairo/src/generate/sources.ts
ericnordelo: After some thought, I don't like removing the current Id computation, since is a unique id that comes from the options, so it identifies the combination, while the incremental name doesn't but comes just from the order. I think we can let the ids as they are, while keeping the names unique. We can remove the id logic from writeGeneratedSources though.

packages/core-cairo/src/test.ts
ericnordelo: Should we remove the testCompile function if we are not going to test the compilation in that test?

packages/core-cairo/test_project/Scarb.toml
ericnordelo: Done!

packages/core-cairo/src/test.ts
ericglau: It may be useful to still generate all combinations (but not compile) as part of the tests, in case any code paths lead to an unexpected error.  This would allow us to notice error conditions earlier in the development cycle.  In this case, we could rename `testCompile` to like `testGenerate`, and in line 34 either generate to the same `test_project` output directory or use a directory within the OS temp dir.  What do you think?

ericnordelo: Updated!

ericglau (approved):
LGTM!

PR #371, opened 2024-06-26
chore: update .gitignore to include .DS_Store

jarrensj:


ericglau:
Thanks for the PR.  However, since this is an OS-specific file and is not specific to the nature of this project, I don't think it belongs in the project's .gitignore.  Consider adding it to a global .gitignore locally instead, for example see https://dev.to/lpheller/quickly-ignore-dsstore-files-globally-ecd

PR #378, opened 2024-08-12
Bump cairo version to 0.15

ericnordelo:


ericnordelo:
@ericglau any idea why tests are still failing after I updated the 0.14.0 version in all the test files?

ericglau:
@ericnordelo You may need to regenerate snapshots, since the library version appears as a comment in the output.
1. cd into the `packages/core-cairo` directory
2. If you’ve changed code generation logic, recompile Typescript files first with `yarn tsc -b` 
3. Update snapshots with `yarn ava --update-snapshots` 
4. Review updated `.md` files, and commit them if they look correct.

ericglau:
@ericnordelo You may need to regenerate snapshots using the `yarn ava --update-snapshots` command and commit the `.snap` files in addition to the `.md` files.  The `.snap` files seem to be something like metadata.

ericglau:
For completeness, I would normally suggest to add a changelog entry in https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/core-cairo/CHANGELOG.md and bump the version in https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/core-cairo/package.json, so we can make a new release of the `@openzeppelin/wizard-cairo` npm package.  This isn't strictly necessary in this case though, since it seems the only change visible to the end user is the library version in the comment.

ericglau (approved):
LGTM, thanks. I added an Unreleased section in the changelog with this change, so we don't lose track of it for the next npm package release.

ericglau:
There is the following error in the Scarb setup step in https://github.com/OpenZeppelin/contracts-wizard/actions/runs/10359200513/job/28675084025?pr=378:
```
Error: There was an error when restoring cache: failed to cache dependencies: unable to hash Scarb.toml file
```

ericnordelo:
> There is the following error in the Scarb setup step in https://github.com/OpenZeppelin/contracts-wizard/actions/runs/10359200513/job/28675084025?pr=378:

This seems to be a warning labeled in the wrong way, that step passes, it looks like the workflow is failing because the compilation takes too much time?

ericglau:
It fails in the `Run software-mansion/setup-scarb@v1` step though, which is before the `Compile contracts` step.

ericnordelo:
> It fails in the Run software-mansion/setup-scarb@v1 step though, which is before the Compile contracts step.

It seems to failed but it passes, the Error is a warning, notice that the setup-scarb step runs successfully

ericglau:
Oh I see what you mean.  In [this previous successful build](https://github.com/OpenZeppelin/contracts-wizard/actions/runs/9796302594/job/27050433219) on `master`, that Error message also exists.  But the compile step in that passing build completed in 6 minutes, while it timed out after 45 minutes here.

If I run a `scarb build` locally with just one contract (keeping only `packages/core-cairo/test_project/src/Contract1.cairo` and deleting the rest), it takes 69 seconds with the latest changes from this PR.  Previously it only took 17 seconds.

ericnordelo:
Eric, I ran the test locally and they pass (after 20 minutes). I have removed the workflow at least temporarily to get the PR merged, later we can figure out what to do about it.

ericglau (approved):

PR #380, opened 2024-08-28
Enable prefilling initial options from URL search parameters

ericglau:
Enables the `name`, `symbol`, and `premint` options to be prefilled from URL search parameters.

Previously the string content after `#` in the URL was treated as the tab (e.g. `erc20` or `erc721`)

With this PR, it splits that string content into further fragments, where a fragment can be just the tab as before, or fragments can be additional `key=value` pairs.

For example: https://wizard.openzeppelin.com/cairo#erc721&name=MyContractName would go to the Cairo language with `tab=erc721` and `name=MyContractName`

### DIFF
--- a/packages/ui/src/App.svelte
+++ b/packages/ui/src/App.svelte
@@ -54,25 +54,16 @@
     $: {
       if (opts) {
         if (!initialValuesSet) {
+          opts.name = initialOpts.name ?? opts.name;
           switch (opts.kind) {
             case 'ERC20':
-              opts.name = initialOpts.name ?? opts.name;
-              opts.symbol = initialOpts.symbol ?? opts.symbol;
               opts.premint = initialOpts.premint ?? opts.premint;
-              break;
             case 'ERC721':
-              opts.name = initialOpts.name ?? opts.name;
               opts.symbol = initialOpts.symbol ?? opts.symbol;
               break;
             case 'ERC1155':
-              opts.name = initialOpts.name ?? opts.name;
-              break;
             case 'Governor':
-              opts.name = initialOpts.name ?? opts.name;
-              break;
             case 'Custom':
-              opts.name = initialOpts.name ?? opts.name;
-              break;
           }
           initialValuesSet = true;
         }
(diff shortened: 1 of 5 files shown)


ericglau:
Examples from preview build:
https://deploy-preview-380--openzeppelin-contracts-wizard.netlify.app/#erc721&name=MyNFT
https://deploy-preview-380--openzeppelin-contracts-wizard.netlify.app/cairo#erc1155&name=My1155Contract

packages/ui/src/App.svelte
```diff
@@ -38,17 +39,43 @@
...
+              break;
+            case 'Governor':
+              opts.name = initialOpts.name ?? opts.name;
+              break;
+            case 'Custom':
+              opts.name = initialOpts.name ?? opts.name;
+              break;
+          }
+          initialValuesSet = true;
+        }
```
ernestognw, line 78: Maybe cleaner for now? Feel free to refactor or ignore, I just noticed name is set many times.

```suggestion
        if (!initialValuesSet) {
          opts.name = initialOpts.name ?? opts.name;
          switch (opts.kind) {
            case 'ERC20':
              opts.premint = initialOpts.premint ?? opts.premint;
            case 'ERC721':
              opts.symbol = initialOpts.symbol ?? opts.symbol;
              break;
            case 'ERC1155':
            case 'Governor':
            case 'Custom':
          }
          initialValuesSet = true;
        }
```

packages/ui/src/cairo/App.svelte
```diff
@@ -31,15 +31,38 @@
...
+              opts.symbol = initialOpts.symbol ?? opts.symbol;
+              break;
+            case 'ERC1155':
+              opts.name = initialOpts.name ?? opts.name;
+              break;
+            case 'Custom':
+              opts.name = initialOpts.name ?? opts.name;
+              break;
+          }
+          initialValuesSet = true;
```
ernestognw, line 64: Similar here.

ernestognw (approved):
LGTM, leaving a couple of nits

PR #386, opened 2024-09-08
Fix cairo hyperlinks

### LINKED ISSUE #385: Broken cairo injected hyperlinks
The new directory structure in Contracts for Cairo, introduced in [v0.15.0](https://github.com/OpenZeppelin/cairo-contracts/releases/tag/v0.15.0), breaks the current hyperlink paths. For example:

```
// Wizard link URL
https://github.com/OpenZeppelin/cairo-contracts/blob/v0.15.0/src/access/ownable.cairo

// Actual URL
https://github.com/OpenZeppelin/cairo-contracts/blob/v0.15.0/packages/access/src/ownable/ownable.cairo
```

andrew-fleming:
Resolves #385.

### DIFF
--- a/packages/ui/src/cairo/inject-hyperlinks.ts
+++ b/packages/ui/src/cairo/inject-hyperlinks.ts
@@ -31,8 +31,8 @@ export function injectHyperlinks(code: string) {
 function removeComponentName(libraryPathSegments: Array<string>) {
   const lastItem = libraryPathSegments[libraryPathSegments.length - 1];
   if (libraryPathSegments.length > 0 && lastItem !== 'interface') {
-    // Replace component name with 'upgradeable'
     if (lastItem === 'UpgradeableComponent') {
+      // Replace component name with 'upgradeable'
       libraryPathSegments.splice(-1, 1, 'upgradeable');
     } else {
       libraryPathSegments.pop();
--- a/packages/ui/src/cairo/inject-hyperlinks.ts
+++ b/packages/ui/src/cairo/inject-hyperlinks.ts
@@ -26,12 +26,10 @@ export function injectHyperlinks(code: string) {
 
 function removeComponentName(libraryPathSegments: Array<string>) {
   const lastItem = libraryPathSegments[libraryPathSegments.length - 1];
-  if (lastItem !== 'interface') {
-    if (lastItem === 'UpgradeableComponent') {
-      // Replace component name with 'upgradeable'
-      libraryPathSegments.splice(-1, 1, 'upgradeable');
-    } else {
-      libraryPathSegments.pop();
-    }
+  if (lastItem === 'UpgradeableComponent') {
+    // Replace component name with 'upgradeable'
+    libraryPathSegments.splice(-1, 1, 'upgradeable');
+  } else {
+    libraryPathSegments.pop();
   }
 }
--- a/packages/ui/src/cairo/inject-hyperlinks.ts
+++ b/packages/ui/src/cairo/inject-hyperlinks.ts
@@ -7,13 +7,31 @@ export function injectHyperlinks(code: string) {
   while (match != null) {
     const [line, libraryPrefix, libraryPath] = match;
     if (line !== undefined && libraryPrefix !== undefined && libraryPath !== undefined) {
-      const githubPrefix = `https://github.com/OpenZeppelin/cairo-contracts/blob/${contractsVersionTag}/src/`;
+      const githubPrefix = `https://github.com/OpenZeppelin/cairo-contracts/blob/${contractsVersionTag}/packages/`;
 
       let libraryPathSegments = libraryPath.split('::');
+      libraryPathSegments.splice(1, 0, 'src');
 
       // Remove the component name
-      if (libraryPathSegments.length > 0 && libraryPathSegments[libraryPathSegments.length - 1] !== 'interface') {
-        libraryPathSegments.pop();
+      let lastItem = libraryPathSegments[libraryPathSegments.length - 1];
+      if (libraryPathSegments.length > 0 && lastItem !== 'interface') {
+        // Replace component name with 'upgradeable'
+        if (lastItem === 'UpgradeableComponent') {
+          libraryPathSegments.splice(-1, 1, 'upgradeable')
+        } else {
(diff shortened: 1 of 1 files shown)


packages/ui/src/cairo/inject-hyperlinks.ts
```diff
@@ -24,4 +26,27 @@ export function injectHyperlinks(code: string) {
     match = importRegex.exec(code);
   }
   return result;
-}
\ No newline at end of file
+}
+
+function removeComponentName(libraryPathSegments: Array<string>) {
+  const lastItem = libraryPathSegments[libraryPathSegments.length - 1];
+  if (libraryPathSegments.length > 0 && lastItem !== 'interface') {
```
immrsd, line 33:
```suggestion
  if (libraryPathSegments.length === 0) { return; }
  const lastItem = libraryPathSegments[libraryPathSegments.length - 1];
  if (lastItem !== 'interface') {
```

packages/ui/src/cairo/inject-hyperlinks.ts
```diff
@@ -24,4 +26,27 @@ export function injectHyperlinks(code: string) {
...
+    if (lastItem === 'UpgradeableComponent') {
+      // Replace component name with 'upgradeable'
+      libraryPathSegments.splice(-1, 1, 'upgradeable');
```
immrsd, line 36: Can you help understand this logic? And is it only applicable for 'upgradable' module?

packages/ui/src/cairo/inject-hyperlinks.ts
```diff
@@ -24,4 +26,27 @@ export function injectHyperlinks(code: string) {
...
+
+/// Duplicate file name as preceding directory in path unless
+/// package name (`libraryPathSegments[0]`) is in `ignoredPackages`.
+function duplicateFileAsDirectoryInPath(libraryPathSegments: Array<string>) {
+  const ignoredPackages = ['security', 'introspection', 'upgrades'];
+  const packageName = libraryPathSegments[0];
+  const lastItem = libraryPathSegments[libraryPathSegments.length - 1];
+  if (lastItem !== undefined && packageName !== undefined && !ignoredPackages.includes(packageName) ) {
+    libraryPathSegments.splice(-1, 0, lastItem);
+  }
```
immrsd, line 51: Could you explain why these specific modules should be ignored?

andrew-fleming: The specified packages in the lib don't include file names with a directory of the same name i.e. `src5/src5.cairo`, `pausable/pausable.cairo`, `upgradeable/upgradeable.cairo`

packages/ui/src/cairo/inject-hyperlinks.ts
andrew-fleming: Yeah, the path for the URL is gathered from the `use` statement and the logic for `UpgradeableComponent` is in `upgradeable.cairo`. Since the `use` statement doesn't include `upgradeable` in the path, we're manually inserting it so the URL points to the correct location. AFAICT it's only applicable to `upgradeable` ATM

ericglau:
The following lead to broken links (select ERC20 and enable everything):
```
    use openzeppelin::token::erc20::extensions::ERC20VotesComponent;
    use openzeppelin::utils::cryptography::nonces::NoncesComponent;
    use openzeppelin::utils::cryptography::snip12::SNIP12Metadata;
```

packages/ui/src/cairo/inject-hyperlinks.ts
ericglau: I feel that these kinds of rules/edge cases may be hard to maintain as additional packages are added in the future.

While the current implementation brings users directly to the source code of the component, which is very useful, I wonder if we should follow a simpler and more generic mapping for reliability.

For example, with import:
```
use openzeppelin::access::ownable::OwnableComponent;
```
We can omit the component name, and just bring users to the `.cairo` file that is named by its path, specifically: https://github.com/OpenZeppelin/cairo-contracts/blob/v0.15.0/packages/access/src/ownable.cairo

(It looks like the only exception is replacing `upgrades` with `upgradeable`, as already addressed in this PR)

From here, users can see that this `.cairo` file exposes the other files that contain the actual components.  It's not too difficult to navigate the sources in GitHub to find it.  This was the intent of the original implementation.

To do this, I think we can simply remove this `duplicateFileAsDirectoryInPath` function.

andrew-fleming: Mm good call, maintainability doesn't seem great...I'm seeing now that ERC20VotesComponent would require even more weird rules to point to the source code.

Simpler sounds better! Will update

packages/ui/src/cairo/inject-hyperlinks.ts
```diff
@@ -24,4 +22,16 @@ export function injectHyperlinks(code: string) {
     match = importRegex.exec(code);
   }
   return result;
-}
\ No newline at end of file
+}
+
+function removeComponentName(libraryPathSegments: Array<string>) {
+  const lastItem = libraryPathSegments[libraryPathSegments.length - 1];
+  if (lastItem !== 'interface') {
```
ericglau, line 29: I think we can remove this `if` condition now.  `lastItem` would be `IUpgradeable` in the case of an interface.  (This condition existed because we previously directly imported the interface path, without component name)

packages/ui/src/cairo/inject-hyperlinks.ts
```diff
@@ -24,4 +22,16 @@ export function injectHyperlinks(code: string) {
...
+    if (lastItem === 'UpgradeableComponent') {
+      // Replace component name with 'upgradeable'
+      libraryPathSegments.splice(-1, 1, 'upgradeable');
+    } else {
+      libraryPathSegments.pop();
+    }
+  }
```
ericglau, line 36:
```suggestion
  const lastItem = libraryPathSegments[libraryPathSegments.length - 1];
  if (lastItem === 'UpgradeableComponent') {
    // Replace component name with 'upgradeable'
    libraryPathSegments.splice(-1, 1, 'upgradeable');
  } else {
    libraryPathSegments.pop();
  }
```

PR #387, opened 2024-09-10
Add cairo account

### LINKED ISSUE #376: Add account option to cairo wizard

andrew-fleming:
Fixes #376.


packages/core-cairo/src/build-generic.ts
```diff
@@ -2,12 +2,14 @@ import { ERC20Options, buildERC20 } from './erc20';
...
 import { CustomOptions, buildCustom } from './custom';
+import { AccountOptions, buildAccount } from './account';
 
 export interface KindedOptions {
   ERC20:    { kind: 'ERC20' }    & ERC20Options;
   ERC721:   { kind: 'ERC721' }   & ERC721Options;
   ERC1155:  { kind: 'ERC1155' }  & ERC1155Options;
-  Custom:  { kind: 'Custom' }  & CustomOptions;
+  Custom:   { kind: 'Custom' }  & CustomOptions;
+  Account:  { kind: 'Account' }  & AccountOptions;
```
ericnordelo, line 12: Does this affect the order?
```suggestion
  Account:  { kind: 'Account' }  & AccountOptions;
  Custom:   { kind: 'Custom' }  & CustomOptions;
```

packages/core-cairo/src/build-generic.ts
```diff
@@ -26,6 +28,9 @@ export function buildGeneric(opts: GenericOptions) {
     case 'Custom':
       return buildCustom(opts);
 
+    case 'Account':
```
ericnordelo, line 31: We should move it on top of Custom.

packages/ui/src/cairo/AccountControls.svelte
```diff
@@ -0,0 +1,81 @@
...
+      <HelpTooltip link="https://docs.starknet.io/architecture-and-concepts/smart-contracts/contract-classes/">
+        Enables the account to declare other contract classes.
+      </HelpTooltip>
+    </label>
+
+    <label class:checked={opts.deploy}>
+      <input type="checkbox" bind:checked={opts.deploy}>
+      Deployable
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts-cairo/accounts#deploying_an_account">
+        Enables the account to counterfactually deploy other contracts.
```
ericnordelo, line 66:
```suggestion
        Enables the account to be counterfactually deployed.
```

packages/ui/src/cairo/AccountControls.svelte
```diff
@@ -0,0 +1,81 @@
...
+</section>
+
+<section class="controls-section">
+  <h1>Account Type</h1>
+  <div class="checkbox-group">
+    <label class:checked={accountType === 'stark'}>
+      <input type="radio" bind:group={opts.type} value="stark">
+      Starknet
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts-cairo/accounts#starknet_account">
+        Starknet account that uses the STARK curve.
```
ericnordelo, line 36:
```suggestion
        Starknet account that uses the STARK curve for signature checking.
```

packages/ui/src/cairo/AccountControls.svelte
```diff
@@ -0,0 +1,81 @@
...
+      </HelpTooltip>
+    </label>
+
+    <label class:checked={accountType === 'eth'}>
+      <input type="radio" bind:group={opts.type} value="eth">
+      Ethereum
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts-cairo/accounts#ethereum_account">
+        Ethereum-flavored account that uses the Secp256k1 curve.
```
ericnordelo, line 44:
```suggestion
        Ethereum-flavored account that uses the Secp256k1 curve for signature checking.
```

ericnordelo (commented):
Great work @andrew-fleming! Left some comments.

ericglau (commented):
This is looking great!  I'm still going through it, but a few initial comments for now.

packages/core-cairo/src/account.ts
```diff
@@ -0,0 +1,189 @@
...
+export const accountOptions = ['stark', 'eth'] as const;
+export type Account = typeof accountOptions[number];
+
+export const defaults: Required<AccountOptions> = {
+  name: 'MyAccount',
+  type: 'stark',
+  declare: true,
+  deploy: true,
+  pubkey: true,
+  access: false,
```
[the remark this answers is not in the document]
andrew-fleming: We can totally do that! I tried not to break the current requirements to support account contracts since they should be handled a little differently. I think the suggestion makes sense though. Will update

packages/core-cairo/src/print.ts
```diff
@@ -6,7 +6,9 @@ import { formatLines, spaceBetween, Lines } from './utils/format-lines';
 import { getSelfArg } from './common-options';
 import { compatibleContractsSemver } from './utils/version';
 
-export function printContract(contract: Contract): string {
+export function printContract(contract: Contract, isAccount?: boolean): string {
```
immrsd, line 9: 1. I think it makes sense to make the `isAccount` parameter required. Then we'll be sure that we have updated all the calls to it throughout the codebase

2. It's not obvious that the 2nd param of `printContract(contract, true)` tells if the contract is an account or not. Also, we may need other flags/params in the future. What do you think on making it a part of the options dictionary?
```suggestion
export function printContract(contract: Contract, options: { isAccount: boolean }): string {
```
This way calls will looks like this
```
printContract(contract, { isAccount: false })
// or
printContract(contract, { isAccount: opts?.kind === 'Account' })
```

immrsd (commented):
Nicely done! I've left a couple of suggestions on improving readability and reliability

packages/core-cairo/src/account.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export function buildAccount(opts: AccountOptions): Contract {
+  const c = new ContractBuilder(opts.name);
+
+  const allOpts = withDefaults(opts);
+
+  if (opts.type === 'stark') {
+    c.addConstructorArgument({ name:'public_key', type:'felt252' });
+    c.addComponent(components.AccountComponent, [{ lit:'public_key' }], true);
+  } else if (opts.type === 'eth') {
```
immrsd, line 61: We'd better go with a switch statement here instead of if/else

packages/core-cairo/src/account.ts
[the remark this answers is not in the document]
andrew-fleming: I ended up creating a `WizardAccountAPI` interface to properly scope the generic `Options` types in the API. This also removes `isAccessControlRequired` from accounts. Let me know what you think

packages/core-cairo/src/common-options.ts
```diff
@@ -5,25 +5,40 @@ import { defaults as infoDefaults } from "./set-info";
...
 
 export const defaults: Required<CommonOptions> = {
+  upgradeable: true,
+  info: infoDefaults,
+} as const;
+
+export const contractDefaults: Required<CommonContractOptions> = {
   access: false,
   upgradeable: true,
   info: infoDefaults,
```
ericglau, line 15:
```suggestion
  ...defaults,
  access: false,
```

packages/core-cairo/src/common-options.ts
```diff
@@ -5,25 +5,40 @@ import { defaults as infoDefaults } from "./set-info";
...
     upgradeable: opts.upgradeable ?? defaults.upgradeable,
     info: opts.info ?? defaults.info,
   };
 }
 
+export function withCommonContractDefaults(opts: CommonContractOptions): Required<CommonContractOptions> {
+  return {
+    access: opts.access ?? contractDefaults.access,
+    upgradeable: opts.upgradeable ?? contractDefaults.upgradeable,
+    info: opts.info ?? contractDefaults.info,
```
ericglau, line 38:
```suggestion
    ...withCommonDefaults(opts),
    access: opts.access ?? contractDefaults.access,
```

packages/core-cairo/src/account.ts
```diff
@@ -0,0 +1,186 @@
...
+import { defaults as commonDefaults } from './common-options';
+import { setAccountUpgradeable } from './set-upgradeable';
+import { setInfo } from './set-info';
+import { defineComponents } from './utils/define-components';
+import { printContract } from './print';
+import { addSRC5Component } from './common-components';
+
+
+export const accountOptions = ['stark', 'eth'] as const;
+export type Account = typeof accountOptions[number];
```
ericglau, line 12: This has a similar name as the AccountOptions interface below, so would be clearer to rename it.
```suggestion
export const accountTypes = ['stark', 'eth'] as const;
export type Account = typeof accountTypes[number];
```

packages/core-cairo/src/print.ts
```diff
@@ -6,7 +6,8 @@ import { formatLines, spaceBetween, Lines } from './utils/format-lines';
 import { getSelfArg } from './common-options';
 import { compatibleContractsSemver } from './utils/version';
 
-export function printContract(contract: Contract): string {
+export function printContract(contract: Contract, options: { isAccount: boolean }): string {
```
ericglau, line 9: This could be simplified by using a default for the options parameter, either like:
```
export function printContract(contract: Contract, options = { isAccount: false }): string {
```
or even just a boolean instead of an object:
```
export function printContract(contract: Contract, isAccount = false): string {
```

That way, for everything that isn't an account, you could just omit the parameter (so no need to pass in `{ isAccount: false })` when calling this from erc20.test.ts etc.)

ericglau (commented):
Could you also add an entry in `packages/core-cairo/CHANGELOG.md`, and add the `account` type in the API readme at `packages/core-cairo/README.md`?

packages/core-cairo/src/print.ts
ericglau: Or even simpler, just put the `account` boolean in the `Contract` interface in `packages/core-cairo/src/contract.ts`, and set that boolean when building the contract in `packages/core-cairo/src/account.ts`

ericglau (approved):
Outstanding 👍

ericnordelo (approved):
LGTM @andrew-fleming.

Let's not forget to rebuild the test_project (by using `yarn run update_scarb_project`), and then compile it locally using scarb, to ensure compilation passes without warnings.

andrew-fleming:
@ericnordelo Done! No issues with compilation

PR #404, opened 2024-11-05
Add stablecoin and real world asset tabs

cairoeth:


ericglau:
Are the "Stablecoin" and "Real-World Assets" tabs intended to be identical?

cairoeth:
> Are the "Stablecoin" and "Real-World Assets" tabs intended to be identical?

Yes

ericglau:
> > Are the "Stablecoin" and "Real-World Assets" tabs intended to be identical?
> 
> Yes -- pointing to same page.

Do we expect these to become different at some point, in a way that is not backwards compatible?  This question relates to semantic versioning of the external API.  If we indicate that these are experimental, then that allows us to change them in the future without bumping the major version.

cairoeth:
> > > Are the "Stablecoin" and "Real-World Assets" tabs intended to be identical?
> > 
> > 
> > Yes -- pointing to same page.
> 
> Do we expect these to become different at some point, in a way that is not backwards compatible? This question relates to semantic versioning of the external API. If we indicate that these are experimental, then that allows us to change them in the future without bumping the major version.

Yeah, these two tabs should be marked/labelled as experimental

packages/core/README.md
```diff
@@ -16,6 +16,8 @@ The following contract types are supported:
 - `erc20`
 - `erc721`
 - `erc1155`
+- `stablecoin`
+- `realWorldAssets`
```
ericglau, line 20: Add a note somewhere in this section that these two are experimental and may be subject to change.

packages/core/src/stablecoin.ts
```diff
@@ -0,0 +1,369 @@
...
+
+export function printStablecoin(opts: StablecoinOptions = defaults): string {
+  return printContract(buildStablecoin(opts));
+}
+
+export function isAccessControlRequired(opts: Partial<StablecoinOptions>): boolean {
+  return opts.mintable || opts.limitations !== false || opts.custodian || opts.pausable || opts.upgradeable === 'uups';
+}
+
+export function buildStablecoin(opts: StablecoinOptions): Contract {
```
ericglau, line 70: Would it make sense to just "extend" erc20.ts in some way to add `limitations` and whatever else is needed, without having to duplicate the code building logic of ERC20?  I don't think this needs to be done for this PR though.

packages/ui/src/StablecoinControls.svelte
```diff
@@ -0,0 +1,167 @@
...
+    <label class="labeled-input">
+      <span class="flex justify-between pr-2">
+        Premint
+        <HelpTooltip>Create an initial amount of tokens for the deployer.</HelpTooltip>
+      </span>
+      <input bind:value={opts.premint} placeholder="0" pattern={premintPattern.source}>
+    </label>
+</section>
+
+<section class="controls-section">
```
ericglau, line 46: After this PR, we should consider whether this can just extend ERC20Controls.svelte in some way, similar to my other comment for the code building.

ericglau (commented):
Initial comments

packages/core/src/generate/stablecoin.ts
```diff
@@ -0,0 +1,31 @@
...
+  symbol: ['MST'],
+  burnable: booleans,
+  pausable: booleans,
+  mintable: booleans,
+  permit: booleans,
+  limitations: booleans,
+  votes: [ ...booleans, ...clockModeOptions ] as const,
+  flashmint: booleans,
+  premint: ['1'],
+  custodian: booleans,
```
ericglau, line 21: If stablecoin.ts has `limitations?: boolean | "allowlist" | "blocklist"`, then this should be `custodian: [ ...booleans, 'allowlist', 'blocklist']`

Or if we change stablecoin.ts to have `limitations?: false | "allowlist" | "blocklist";` as per my other comment, then this should be `custodian: [ false, 'allowlist', 'blocklist']`

packages/core/src/stablecoin.ts
```diff
@@ -0,0 +1,369 @@
...
+
+export interface StablecoinOptions extends CommonOptions {
+  name: string;
+  symbol: string;
+  burnable?: boolean;
+  pausable?: boolean;
+  premint?: string;
+  mintable?: boolean;
+  permit?: boolean;
+  limitations?: boolean | "allowlist" | "blocklist";
```
ericglau, line 19: Is there any need to support `true`, or should this be reduced to just be 
```suggestion
  limitations?: false | "allowlist" | "blocklist";
```
This would allow the logic in `addLimitations` to be simplified.

packages/ui/src/App.svelte
```diff
@@ -183,6 +188,12 @@
         <button class:selected={tab === 'ERC1155'} on:click={() => tab = 'ERC1155'}>
           ERC1155
         </button>
+        <button class:selected={tab === 'Stablecoin'} on:click={() => tab = 'Stablecoin'}>
+          Stablecoin
+        </button>
+        <button class:selected={tab === 'RealWorldAssets'} on:click={() => tab = 'RealWorldAssets'}>
+          Real-World Assets
+        </button>
```
ericglau, line 196: The difference in singular vs. plural for "Stablecoin" vs. "Real-World Assets" is somewhat jarring.  Is this difference intentional?

ericglau: I see that this can make sense if the contract itself represents a "Stablecoin" with many units, whereas each unit is an asset (so the contract itself represents "Real-World Assets").  Feel free to resolve this if this was the intent.

cairoeth: Yes intended

packages/core/src/stablecoin.ts
```diff
@@ -0,0 +1,369 @@
...
+function addMintable(c: ContractBuilder, access: Access) {
+  requireAccessControl(c, functions.mint, access, 'MINTER', 'minter');
+  c.addFunctionCode('_mint(to, amount);', functions.mint);
+}
+
+function addLimitations(c: ContractBuilder, access: Access, mode: boolean | 'allowlist' | 'blocklist') {
+  const type = mode === 'allowlist';
+  const ERC20Limitation = {
+    name: type ? 'ERC20Allowlist' : 'ERC20Blocklist',
+    path: `@openzeppelin/community-contracts/contracts/token/ERC20/extensions/${type ? 'ERC20Allowlist' : 'ERC20Blocklist'}.sol`,
```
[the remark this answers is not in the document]
cairoeth: I don't think we can do `@openzeppelin/community-contracts/token/...` as the community repo doesn't define the package inside the contracts directory, whereas the main repo does: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/package.json

packages/core/src/stablecoin.ts
```diff
@@ -0,0 +1,369 @@
...
+    const clockMode = allOpts.votes === true ? clockModeDefault : allOpts.votes;
+    addVotes(c, clockMode);
+  }
+
+  if (allOpts.flashmint) {
+    addFlashMint(c);
+  }
+
+  setAccessControl(c, access);
+  // setUpgradeable(c, upgradeable, access);
```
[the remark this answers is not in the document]
cairoeth: https://github.com/OpenZeppelin/contracts-wizard/pull/404/commits/7d26a46e7d6c11887a529bf65db896733bf050a9

packages/core/src/stablecoin.ts
```diff
@@ -0,0 +1,369 @@
...
+        c.addConstructorCode(`_grantRole(${roleId}, ${roleOwner});`);
+      }
+      c.setFunctionBody([`return hasRole(CUSTODIAN_ROLE, user);`], functions._isCustodian);
+      break;
+    }
+    case 'managed': {
+      const logic = [
+        `(bool immediate,) = AuthorityUtils.canCallWithDelay(authority(), user, address(this), bytes4(_msgData()[0:4]));`,
+        `if (!immediate) revert AccessManagedUnauthorized(user);`,
+        `return true;`
```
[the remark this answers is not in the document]
cairoeth: agreed https://github.com/OpenZeppelin/contracts-wizard/pull/404/commits/7d26a46e7d6c11887a529bf65db896733bf050a9

packages/core/README.md
cairoeth: https://github.com/OpenZeppelin/contracts-wizard/pull/404/commits/cdd5198f56a6ef0df5f1b9f23dd006b229a39bfd

packages/core/src/stablecoin.ts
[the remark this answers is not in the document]
ericglau: We do have a disclaimer that the community contracts repo can have breaking changes, so I think we can leave this for now and change it when needed.

packages/core/src/stablecoin.ts
```diff
@@ -0,0 +1,372 @@
...
+        c.addConstructorArgument({type: 'address', name: roleOwner});
+        c.addConstructorCode(`_grantRole(${roleId}, ${roleOwner});`);
+      }
+      c.setFunctionBody([`return hasRole(CUSTODIAN_ROLE, user);`], functions._isCustodian);
+      break;
+    }
+    case 'managed': {
+      const logic = [
+        `(bool immediate,) = AuthorityUtils.canCallWithDelay(authority(), user, address(this), bytes4(_msgData()[0:4]));`,
+        `if (!immediate) return false;`,
```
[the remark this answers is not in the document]
cairoeth: yup good one

ericglau:
Some notes:

1. Fixed some issues with the Real-World Assets tab -- the only way I could get it working was to add a separate svelte file for it, instead of reusing StablecoinControls.svelte to avoid conflicts with the `kind` option.  In a future PR we could try to make one of those import the other as well.

2. Since the community repo is on GitHub only, download packages for Hardhat and Foundry currently do not handle it, so I've disabled those for these new contract kinds for now.

3. Remix also has an issue due with compiling.  Even if I change the code in Remix to import with the following format:
```
import "@openzeppelin/contracts@5.1.0/token/ERC20/ERC20.sol";
import "https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Custodian.sol";
import "@openzeppelin/contracts@5.1.0/access/Ownable.sol";
```
the GitHub repo's ERC20Custodian.sol imports a version of ERC20 which conflicts with the one from `@openzeppelin/contracts@5.1.0`.
So I've disabled "Open in Remix" for these new contract kinds as well.

ericglau:
At this point I think the only remaining item is to mark these tabs as experimental in the UI.

packages/ui/src/App.svelte
```diff
@@ -183,6 +189,12 @@
         <button class:selected={tab === 'ERC1155'} on:click={() => tab = 'ERC1155'}>
           ERC1155
         </button>
+        <button class:selected={tab === 'Stablecoin'} on:click={() => tab = 'Stablecoin'}>
+          Stablecoin
```
ericglau, line 193:
```suggestion
          Stablecoin*
```

packages/ui/src/App.svelte
```diff
@@ -183,6 +189,12 @@
...
+        </button>
+        <button class:selected={tab === 'RealWorldAsset'} on:click={() => tab = 'RealWorldAsset'}>
+          Real-World Asset
```
ericglau, line 196:
```suggestion
          Real-World Asset*
```

ericglau (approved):
LGTM, thanks!

PR #405, opened 2024-11-15
Bump Cairo library to 0.19.0

immrsd:
There were no breaking changes in `v0.19.0` that had to be supported.

Executed `yarn run update_scarb_project` and successfully build the test target.
image


packages/core-cairo/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.18.0 (2024-11-15)
```
ericglau, line 3: Since this does not have code changes, I don't think we need to release an update to the package at this point.
```suggestion
## Unreleased
```

ericglau (approved):
Just one comment. The rest LGTM, thanks!

packages/core-cairo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openzeppelin/wizard-cairo",
-  "version": "0.17.0",
+  "version": "0.18.0",
```
ericglau, line 3: We could revert this and bump when a release is needed
```suggestion
  "version": "0.17.0",
```

andrew-fleming (approved):
LGTM!

packages/core-cairo/CHANGELOG.md
ericglau: I think that is ok, but it won't be practical to always ensure the versions are aligned.

PR #416, opened 2024-12-02
Add support to defender deploy

MCarlomagno:
## Description

Adds a modal that connects the wizard with the [Defender Deploy Plugin](https://github.com/OpenZeppelin/defender-remix-plugin) (currently only supported in remix).

The modal contains an Iframe that points to the [live version of the remix plugin](https://defender-deploy-wizard.netlify.app/).


packages/ui/src/standalone.css
```diff
@@ -111,9 +111,9 @@ body {
   margin: -12rem auto 0;
 }
 .wizard {
-  background-color: #f9fafb;
+  /*background-color: #f9fafb;
   box-shadow: 0px 10px 40px rgb(108 133 171 / 20%);
-  border-radius: 1rem;
+   border-radius: 1rem; */
```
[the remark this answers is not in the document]
ericglau: IMO we should remove all of the commented styles throughout this PR if they are no longer relevant.

packages/ui/src/App.svelte
```diff
@@ -313,17 +330,163 @@
...
+      <pre class="flex flex-col grow basis-0 overflow-auto">
+        <code class="hljs grow overflow-auto p-4">{@html highlightedCode}</code>
+      </pre>
+      <DefenderDeployModal isOpen={showDeployModal} />
     </div>
   </div>
 </div>
 
 <style lang="postcss">
+ /* start of the magic */
```
ericglau, line 390: Can we put a more detailed comment on what exactly this section does?

packages/ui/src/App.svelte
```diff
@@ -16,23 +16,37 @@
...
     import { injectHyperlinks } from './utils/inject-hyperlinks';
     import { InitialOptions } from './initial-options';
+    import { postMessageToIframe } from './post-message';
+    import { debouncer } from './utils/helpers';
 
     const dispatch = createEventDispatcher();
 
+    // listens to contract changes and posts them to the defender deploy iframe.
+    // we debounce the call to avoid sending too many messages while the user is editing.
+    const debouncedPostMessage = debouncer((contract) => {
```
ericglau, line 43: While this is waiting for the debounce timeout, there is no indication to the user that the iframe is waiting to be updated.  The iframe just gets updated 600 ms later, which "feels" like lag.  

@makiopen Any suggestions on how to add some indication that the iframe is pending an update?

ericglau (commented):
Looking good so far, thanks!  Just left some initial comments.

I also notice if I close the iframe and re-open it, the API key needs to be entered again.  Do we have a way to persist that information for the current session?

packages/ui/src/App.svelte
MCarlomagno: 🧙‍♂️🪄

packages/ui/src/App.svelte
MCarlomagno: We can display some loading state with some "compiling contract" message. We would need to make another post message (instantly) to the iframe to activate the loading state until the debounced message arrives

packages/ui/src/App.svelte
makiopen: 🪄

MCarlomagno:
> I also notice if I close the iframe and re-open it, the API key needs to be entered again. Do we have a way to persist that information just for the current session? Or avoid truly closing it, but just hide it.

Thanks! I ended up hiding the iframe instead of conditionally loading the component, a bit hacky but works

ericglau:
The AI Assistant icon also seems missing, not sure where it went.  Perhaps due to the formatting changes.

makiopen:
some z-index issues, brought it back

ericglau (approved):
Added title (hover) text for "Copy to Clipboard" and "Open in Remix" since they are now icon-only and it may not be obvious what they do.

LGTM, thanks!

packages/ui/src/standalone.css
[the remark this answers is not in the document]
makiopen: removed

packages/ui/src/App.svelte
makiopen: even a spinner would be enough tbh

PR #417, opened 2024-12-03
Add Governor

ericnordelo:
This PR adds Governor to Cairo, and also refactors some code. Summary below:

**Added**:

- `section` member to `Impl` and `ImplementedTrait` interfaces allowing to group in sections under a comment.
- `embed` member to `Impl` interface replacing the `internalImpl` mechanism when adding components since there could be multiple internal impls as shown in Governor.
- `addUseClause` logic replacing the way imports were handled, allowing us to group imports.
- Governor tab.

**This PR is still missing:**

- To refactor the existing tests and add new ones.
- Use macro for pow instead of the 2e18 notation (this macro must be implemented first).

**With this said, it is ready for a first round of reviews since the main logic is finished.**


ericnordelo:
@ericglau having the deploy preview for this first round of reviews would be very helpful.

packages/ui/src/cairo/GovernorControls.svelte
```diff
@@ -0,0 +1,231 @@
...
+      </HelpTooltip>
+    </label>
+  </div>
+</section>
+
+<section class="controls-section">
+  <h1>
+    <!-- svelte-ignore a11y-label-has-associated-control -->
+    <label class="flex justify-between items-center tooltip-container pr-2">
+      <span>SNIP12 Metadata</span>
```
andrew-fleming, line 207: Is it me or does the `12` look too big on the page?

andrew-fleming (commented):
Great start, Eric! I left some comments and suggestions

packages/core-cairo/src/print.ts
```diff
@@ -132,23 +170,53 @@ function printImplementedTraits(contract: Contract): Lines[] {
...
-  for (const trait of sortedTraits) {
-    const implLines = [];
-    implLines.push(...trait.tags.map(t => `#[${t}]`));
-    implLines.push(`impl ${trait.name} of ${trait.of} {`);
-
-    const superVars = withSemicolons(
-      trait.superVariables.map(v => `const ${v.name}: ${v.type} = ${v.value}`)
-    );
-    implLines.push(superVars);
+  // group by section
```
ericglau, line 173: In the code above this, it first sorts by `priority`.  That was originally done to give certain sections such as SNIP12 metadata more visibility by placing it above other implemented traits in ERC20 and ERC721.

With the addition of `section`, there seems to be two layers of sorting.  It's not very clear what is the intended order of these two sorts.

Would it be sufficient to just rely on `section` and remove the `priority` order?  That would change the order of SNIP12 metadata in ERC20/721 though, but perhaps adding the section header would help maintain visibility (as in @andrew-fleming's comment in https://github.com/OpenZeppelin/contracts-wizard/pull/417/files#r1872434347)

packages/core-cairo/src/print.ts
```diff
@@ -59,7 +62,26 @@ function sortImports(contract: Contract): string[] {
...
+  for (const constant of contract.constants) {
+    // inlineComment is optional, default to false
+    let inlineComment = constant.inlineComment ?? false;
+    let commented = !!constant.comment;
+
+    if (commented && !inlineComment) {
+      lines.push(`// ${constant.comment}`);
+      lines.push(`const ${constant.name}: ${constant.type} = ${constant.value};`);
+    } else if (commented) {
+      lines.push(`const ${constant.name}: ${constant.type} = ${constant.value}; /* ${constant.comment} */`);
```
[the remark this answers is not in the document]
ericnordelo: > We should make the same adjustment /* */ => // for line 291

We can't update it for this case since this is an inline comment that may contain code after it.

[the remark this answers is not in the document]
ericnordelo: We can change it for constants, but I think it looks better if we use this notation for inline comments in general.

packages/core-cairo/src/governor.ts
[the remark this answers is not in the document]
ericnordelo: Good catch. Added both InternalExtendedImpl and DefaultConfig. I don't think we should add the  option for configuring DEFAULT_PARAMS on wizard though, since the format is `Span<felt252>` (most probably a serialized struct).

packages/ui/src/cairo/GovernorControls.svelte
ericnordelo: It kind of does, but I think it is just the font used, not sure if we can do something about it.

packages/core-cairo/src/print.ts
```diff
@@ -132,23 +170,53 @@ function printImplementedTraits(contract: Contract): Lines[] {
...
+  return spaceBetween(...sections);
+}
 
-    implLines.push('}');
-    impls.push(implLines);
+function printImplementedTraitsSection(section: string, impls: ImplementedTrait[]) {
+  const lines = [];
+  const isDefaultSection = section === DEFAULT_SECTION;
+  if (!isDefaultSection) {
+    lines.push('//');
```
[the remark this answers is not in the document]
ericnordelo: It is intentional. It is the format we are following in our presets. It could be improved probably, but I wanted to be at least consistent.

packages/core-cairo/src/print.ts
ericnordelo: Priority now allows us to sort inside sections, which is nice. I vote we keep it, and maybe refactor it later into a better descriptive way that may allow us to define the order of sections (outside the default alphabetical sorting).

packages/core-cairo/src/print.ts
```diff
@@ -132,23 +170,53 @@ function printImplementedTraits(contract: Contract): Lines[] {
...
+    lines.push(`// ${section}`);
+    lines.push('//');
```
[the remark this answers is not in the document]
ericnordelo: Done!

ericnordelo:
Added an `addUseClause` logic replacing the `addStandaloneImport`  one. This allows us to optionally group imports.

packages/core-cairo/src/governor.ts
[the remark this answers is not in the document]
andrew-fleming: > I don't think we should add the option for configuring DEFAULT_PARAMS on wizard though, since the format is Span<felt252> (most probably a serialized struct).

Hmm I'm wondering if we should say more about `DEFAULT_PARAMS` in the docs (or in a future guide). IMO it's not obvious what this should be and how it should be used. I agree though in that it's prob not an option we need in wizard

packages/core-cairo/src/contract.ts
```diff
@@ -127,11 +145,19 @@ export class ContractBuilder implements Contract {
     return this.interfaceFlagsSet;
   }
 
-  addStandaloneImport(fullyQualified: string): void {
-    this.standaloneImportsSet.add(fullyQualified);
+  addUseClause(containerPath: string, name: string, options?: { groupable?: boolean, alias?: string }): void {
+    // groupable defaults to true
+    const groupable = options?.groupable ?? true;
+    const alias = options?.alias ?? '';
+    let uniqueName = alias.length > 0 ? alias : name;
```
ericglau, line 152:
```suggestion
    const uniqueName = alias.length > 0 ? alias : name;
```

packages/core-cairo/src/print.ts
```diff
@@ -41,25 +47,123 @@ function printSuperVariables(contract: Contract): string[] {
...
 
-function sortImports(contract: Contract): string[] {
-  const componentImports = contract.components.flatMap(c => `${c.path}::${c.name}`);
-  const allImports = componentImports.concat(contract.standaloneImports);
-  const superVars = contract.superVariables;
-  if (superVars.length === 1) {
-    allImports.push(`super::${superVars[0]!.name}`);
-  } else if (superVars.length > 1) {
-    allImports.push(`super::{${superVars.map(v => v.name).join(', ')}}`);
+function getLinesFromUseClausesGroup(group: UseClause[], groupName: string): Lines[] {
```
ericglau, line 68: It could be slightly simpler to set `alias` to `undefined` in the caller and check its existence, instead of checking for empty string in multiple places.

ericnordelo: The rationale of checking for an empty string, is that if we check for undefined, and for some reason an empty string gets assigned, the print function would print "name as []". This can still be avoided of course with some logic, but I felt is simpler like this.

packages/core-cairo/src/print.ts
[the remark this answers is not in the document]
ericnordelo: Oh, you are right. It is not implemented 😢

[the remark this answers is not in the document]
ericnordelo: I can compile with it, but the language server doesn't work well with it, I will remove it for now, seems it looks unstable 😢

ericglau (approved):
Looks great, thanks!

packages/ui/src/cairo/GovernorControls.svelte
```diff
@@ -0,0 +1,231 @@
...
+    </label>
+  </div>
+
+  <label class="labeled-input">
+    <span class="flex justify-between pr-2">
+      Proposal Threshold
+      <HelpTooltip>Minimum number of votes an account must have to create a proposal.</HelpTooltip>
+    </span>
+    <input bind:value={opts.proposalThreshold} placeholder="0" use:error={errors?.proposalThreshold}>
+  </label>
```
andrew-fleming, line 92: Small thing: do you think it's worth including an error when the threshold exceeds u256? It's pretty easy for this to happen by setting token decimals to a high number

andrew-fleming (approved):
Improvements look great! I just left a non-blocking question

packages/ui/src/cairo/GovernorControls.svelte
ericnordelo: It could be nice, but I don't think it is that helpful in practice, since that shouldn't happen, and if it happens it is a compilation error anyway.

ericglau: Same should apply for Quorum.  See [helper functions](https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/core-cairo/src/utils/convert-strings.ts) that could be used.

ericglau:
@ericnordelo Can you add to the changelog entries under 0.20.0?  I haven't published that version yet so we might as well release these together.  We will be looking into a more formal release process going forward.

packages/ui/src/cairo/GovernorControls.svelte
andrew-fleming: I agree that it really shouldn't happen and yeah, it'll be a compilation error. IMO it'd be better to handle that here though. We do already check that erc20's premint is in range, so it feels a little inconsistent to not check this as well, no?

andrew-fleming:
> I'm having issues building the test_project locally (the process exits after a couple of hours of running). Do you guys have a strategy for this? I'm on an M3 MacBook Pro 😢

Ahh hours!? I unfortunately don't have a strat for this :( I'm on an m2 and the last PR took "only" 30 min to compile

ericglau:
> I'm having issues building the test_project locally (the process exits after a couple of hours of running).

Compile time seems non-linearly increasing for each added contract.  Perhaps try removing the other contract types other than Governor when testing it locally.  That might be sufficient for now.

ericglau:
We could also change the script to generate sources for each contract kind separately.  There there would be like 5 projects that need to be compiled.  That should cut down each project's compile time though.  (Does not need to be in this PR)

packages/ui/src/cairo/GovernorControls.svelte
ericnordelo: Fair enough. Added!

ericglau (approved):

PR #411, opened 2024-11-19
Use named imports for Solidity

ericglau:
Fixes #352

PR #420, opened 2024-12-13
Fixing edge case when import files with same contract name at least two times

Ratimon:
There is an edge case as following:

Suppose I would like to accomplish following code block:

```
pragma solidity ^0.8.20;

import "@redprint-core/cannon/libraries/CannonErrors.sol";
import "@redprint-core/cannon/libraries/CannonTypes.sol";
```

and I use :

```
    const CannonErrors = {
        name: '',
        path: '@redprint-core/cannon/libraries/CannonErrors.sol',
    };
    c.addImportOnly(CannonErrors);
    const CannonTypes = {
        name: '',
        path: '@redprint-core/cannon/libraries/CannonTypes.sol',
    };    
    c.addImportOnly(CannonTypes);

```

However, it will generate only one path with excessive  `{} from` as following  : 

```
/ SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

import {} from "@redprint-core/cannon/libraries/CannonErrors.sol";

```

Adding additional state of  ` readonly modules: ImportContract[] = [];` with modified `contract.ts` and `print.ts` 's functions as PR will solve this:

Please suggest if this is not appropriate, I am happy to make another PR.

Reference to  : [my repo](https://github.com/Ratimon/redprint-wizard/commit/4207f911fcfacd0a157a892e226fbc15ae0458b5)

Regards

### DIFF
--- a/packages/core/src/contract.ts
+++ b/packages/core/src/contract.ts
@@ -5,6 +5,7 @@ export interface Contract {
   license: string;
   parents: Parent[];
   natspecTags: NatspecTag[];
+  modules: ImportContract[];
   imports: ImportContract[];
   functions: ContractFunction[];
   constructorCode: string[];
@@ -80,6 +81,7 @@ export class ContractBuilder implements Contract {
   upgradeable = false;
 
   readonly using: Using[] = [];
+  readonly modules: ImportContract[] = [];
   readonly natspecTags: NatspecTag[] = [];
 
   readonly constructorArgs: FunctionArgument[] = [];
@@ -109,6 +111,7 @@ export class ContractBuilder implements Contract {
     return [
       ...[...this.parentMap.values()].map(p => p.contract),
       ...this.using.map(u => u.library),
+      ...this.modules,
     ];
   }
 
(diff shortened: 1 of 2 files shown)


ericglau:
Hi, thanks for the PR.  However, this code is internal logic and not part of the Contracts Wizard's external API.  We currently do not have a need for importing entire Solidity files in the output that the Contracts Wizard generates, and this was an intentional change to use named imports in https://github.com/OpenZeppelin/contracts-wizard/pull/411.  As such, I don't think it is appropriate to add code paths that we do not use.

The recommended best practice is to use named imports, and you can see this pattern in use in OpenZeppelin Contracts, and this is also recommended in style guides such as [Coinbase's style guide](https://github.com/coinbase/solidity-style-guide?tab=readme-ov-file#6-imports).  

As your project uses a fork of this code, it looks like you already have the changes that you need in your repo already, so it does not depend on this PR.  I will close this PR, but feel free to continue the conversation if needed.  Thanks.

Ratimon:
Thanks for the review!!

PR #425, opened 2024-12-24
Add Vesting

immrsd:


packages/core-cairo/src/utils/duration.ts
```diff
@@ -14,7 +14,7 @@ const secondsForUnit = { second, minute, hour, day, week, month, year };
 export function durationToTimestamp(duration: string): number {
   const match = duration.trim().match(durationPattern);
 
-  if (!match) {
+  if (!match || match.length < 2) {
```
andrew-fleming, line 17: Are there cases where a happy `match` also has `match.length < 2`?

andrew-fleming (commented):
Very nice PR, @immrsd! I left a few questions and comments

packages/core-cairo/src/utils/duration.ts
immrsd: No, since later on we access the 1st and the 2nd items of the match array in unsafe manner. Either way it'll result in a runtime exception, but it's better to have a straightforward error than index-out-of-bounds one

image

packages/core-cairo/src/vesting.ts
```diff
@@ -0,0 +1,223 @@
...
+import { setAccessControl } from './set-access-control';
+import { setUpgradeable } from './set-upgradeable';
+import { Info, setInfo } from './set-info';
+import { defineComponents } from './utils/define-components';
+import { printContract } from './print';
+import { OptionsError } from './error';
+import { durationToTimestamp } from './utils/duration';
+import { toUint, validateUint } from './utils/convert-strings';
+
+export type VestingSchedule = 'linear' | 'custom';
```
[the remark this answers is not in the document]
immrsd: Actually, I don't reckon it's worth it. The steps schedule is mostly for demonstration purposes, it's kinda naive and I expect it to be used almost never or very rarely

ericnordelo (approved):
The PR and generated code look good, but I wonder if there's anything we can do to make the date input UI compatible with the rest of the inputs, since it looks kind of broken cc @ericglau

ericglau:
Applied styling to the date input. cc @ericnordelo 

Screenshot 2025-01-07 at 3 07 15 PM

ericglau (approved):
LGTM, thanks!

immrsd:
Results of test project compilation with only Vesting being enabled:

image

PR #426, opened 2025-01-13
Remove `isAccessControlRequired` from specific contract APIs, update copyright year

ericglau:
- Removes the `isAccessControlRequired` function from the APIs for `governor` and `vesting` by separating their interfaces from contracts that have access control options.
  - Although `account`, `governor`, and `vesting` have their own ways of handling access control, they don't require this function because it's only purpose was to allow the caller to inspect the default behavior of the `access` option.
- Update Cairo API readme.
- Updates copyright year in preparation for publishing.


ericnordelo (approved):
LGTM!

PR #427, opened 2025-01-16
Add Cairo links when multiple components are imported

### LINKED ISSUE #423: Cairo: Add hyperlinks to imports that have multiple items
Hyperlinks for contract imports are injected for `use` lines, which previously imported single items only.

In https://github.com/OpenZeppelin/contracts-wizard/pull/417, this changed multiple items from the same package to be imported in the same line (or multiple lines), which our hyperlink injection does not detect.

Examples:
```
use openzeppelin::governance::governor::{DefaultConfig, GovernorComponent};
use openzeppelin::governance::governor::extensions::{
	GovernorCountingSimpleComponent, GovernorSettingsComponent,
	GovernorTimelockExecutionComponent, GovernorVotesQuorumFractionComponent
};
```

We should determine a scheme for detecting these imports and where it should link to.

Perhaps for the above examples, we can just link to the prefixed package directories on GitHub: https://github.com/OpenZeppelin/cairo-contracts/tree/main/packages/governance/src/governor
https://github.com/OpenZeppelin/cairo-contracts/tree/main/packages/governance/src/governor/extensions

ericglau:
- If multiple components are imported, link to the parent .cairo file with the same name as the last path segment
- If a single component is imported, link to the file containing the component (if defined in the known mappings), otherwise link to the parent .cairo file with the same name as the last path segment

Fixes #423

### DIFF
--- a/packages/ui/src/cairo/inject-hyperlinks.ts
+++ b/packages/ui/src/cairo/inject-hyperlinks.ts
@@ -1,22 +1,32 @@
 import { contractsVersionTag } from "@openzeppelin/wizard-cairo/src";
 
 export function injectHyperlinks(code: string) {
-  const importRegex = /use<\/span> (openzeppelin)::([^\s]*);/g
+  const importRegex = /use<\/span> (openzeppelin)::([^A-Z]*)(::[a-zA-Z0-9]+|::{)/g
   let result = code;
   let match = importRegex.exec(code);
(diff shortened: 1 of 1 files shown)


ericnordelo (approved):
LGTM! We should add more entries to the components mapping, but we can do that in a separate PR.

PR #430, opened 2025-01-17
Remove unused dependencies

ericglau:
- Removes `array.prototype.flatmap` and `array.prototype.flat` since `flat` and `flatMap` are now supported in current LTS versions of Node JS.
- Removes unused `bn.js` and `ethereum-cryptography` from Cairo Wizard package.

PR #436, opened 2025-01-31
Add ERC20 Cross-Chain Bridging, SuperchainERC20

ericglau:
Adds "Cross-Chain Bridging" feature for ERC20 (and Stablecoin and RWA tabs), with the selectable options being "Custom" or "SuperchainERC20".

These are based on ERC20Bridgeable from OpenZeppelin Community Contracts, where:
- "Custom" is intended to be functionally the same as [ERC-7802's reference implementation](https://github.com/ethereum/ERCs/blob/bf1ebe7d946d69c90c9f443df658fd37521b69f7/ERCS/erc-7802.md#reference-implementation), but is restricted differently when using Roles or Managed access control.
- "SuperchainERC20" is intended to be functionally the same as [Optimism's SuperchainERC20 implementation](https://docs.optimism.io/stack/interop/superchain-erc20).

[Wizard UI Preview](https://deploy-preview-436--openzeppelin-contracts-wizard.netlify.app/)


packages/core/solidity/src/erc20.test.ts
```diff
@@ -92,6 +92,63 @@ testERC20('erc20 flashmint', {
   flashmint: true,
 });
 
+testERC20('erc20 bridgeable', {
+  bridgeable: true,
+});
+
+testERC20('erc20 bridgeable ownable', {
+  bridgeable: true,
+});
```
Amxx, line 101: that is missing ownable, right ?

packages/core/solidity/src/erc20.ts
```diff
@@ -204,6 +213,72 @@ function addFlashMint(c: ContractBuilder) {
   });
 }
 
+function addBridgeable(c: ContractBuilder, bridgeable: boolean | 'superchain', upgradeable: false | 'transparent' | 'uups', access: Access) {
+  if (bridgeable === false) {
+    return;
+  }
```
Amxx, line 219: Would that work ?
```suggestion
function addBridgeable(c: ContractBuilder, bridgeable: true | 'superchain', upgradeable: false | 'transparent' | 'uups', access: Access) {
```

packages/core/solidity/src/erc20.ts
```diff
@@ -204,6 +213,72 @@ function addFlashMint(c: ContractBuilder) {
...
+        });
+        const logic = [
+          `(bool immediate,) = AuthorityUtils.canCallWithDelay(authority(), caller, address(this), bytes4(_msgData()[0:4]));`,
+          `if (!immediate) revert Unauthorized();`
+        ]
+        c.setFunctionBody(logic, functions._checkTokenBridge, 'view');
+        break;
+      }
+    }
+  }
```
Amxx, line 278:
```suggestion
  switch (bridgeable) {
    case 'superchain':
      c.addVariable('address internal constant SUPERCHAIN_TOKEN_BRIDGE = 0x4200000000000000000000000000000000000028;');
      c.setFunctionBody(['if (caller != SUPERCHAIN_TOKEN_BRIDGE) revert Unauthorized();'], functions._checkTokenBridge, 'pure');
      break;
    case true: {
      switch (access) {
        case false:
        case 'ownable': {
          const addedImmutable = c.addVariable(`address public immutable TOKEN_BRIDGE;`);
          if (addedImmutable) {
            c.addConstructorArgument({type: 'address', name: 'tokenBridge'});
            c.addConstructorCode(`require(tokenBridge != address(0), "Invalid TOKEN_BRIDGE address");`);
            c.addConstructorCode(`TOKEN_BRIDGE = tokenBridge;`);
          }
          c.setFunctionBody([`if (caller != TOKEN_BRIDGE) revert Unauthorized();`], functions._checkTokenBridge, 'view');
          break;
        }
        case 'roles': {
          setAccessControl(c, access);
          const roleOwner = 'tokenBridge';
          const roleId = 'TOKEN_BRIDGE_ROLE';
          const addedConstant = c.addVariable(`bytes32 public constant ${roleId} = keccak256("${roleId}");`);
          if (addedConstant) {
            c.addConstructorArgument({type: 'address', name: roleOwner});
            c.addConstructorCode(`_grantRole(${roleId}, ${roleOwner});`);
          }
          c.setFunctionBody([`if (!hasRole(${roleId}, caller)) revert Unauthorized();`], functions._checkTokenBridge, 'view');
          break;
        }
        case 'managed': {
          setAccessControl(c, access);
          c.addImportOnly({
            name: 'AuthorityUtils',
            path: `@openzeppelin/contracts/access/manager/AuthorityUtils.sol`,
          });
          const logic = [
            `(bool immediate,) = AuthorityUtils.canCallWithDelay(authority(), caller, address(this), bytes4(_msgData()[0:4]));`,
            `if (!immediate) revert Unauthorized();`
          ]
          c.setFunctionBody(logic, functions._checkTokenBridge, 'view');
          break;
        }
      }
      break;
    }
  }
```

ericglau:

ericglau:
@SocketSecurity ignore npm/undici@5.28.4

Unrelated to this PR, see https://github.com/OpenZeppelin/contracts-wizard/pull/430#issuecomment-2602776853

ericglau:
Added a Chain ID field under Premint, which only appears and is only required when both Premint has a value and Cross-Chain Bridging is enabled.

ernestognw (approved):
Looking good. I would suggest setting the default chainId `1` instead of the current input focus.

Captura de pantalla 2025-02-26 a la(s) 9 43 10 a m

The reason is that the copy button on top is still enabled, allowing to copy the code without the premint option.

ericglau:
@ernestognw I changed it to disable the action buttons (Copy etc.) and make the code non-selectable when there are errors in the input.
Screenshot 2025-02-26 at 1 47 01 PM

I'm not sure if we should set the default chainId to `1`, in case users want to use a ERC-7802 bridge that not available on Ethereum L1.

For example, if using the predeployed SuperchainTokenBridge, the ERC-7802 mints and burns would only be on Superchain networks and not on Ethereum.  I think we should require the chainId in this case, since users may want to premint on something like OP Mainnet instead.

ericglau:

PR #438, opened 2025-02-04
Move Solidity and Cairo packages into subfolders

ericglau:
Moves the folders with code generation logic into separately named subfolders under `core`, to allow for expansion with additional languages/frameworks going forward.
```
packages/core -> packages/core/solidity
packages/core-cairo -> packages/core/cairo
```

Reorganizes UI folders to be as follows:
```
ui/src/solidity
ui/src/cairo
ui/src/common
ui/src (top level files only)
```


ernestognw (approved):
Looks good but it's the kind of PR that might be missing something given its size.
I see everything is passing so it should be fine.

PR #445, opened 2025-02-18
Add Erc1155 Template

0xNeshi:
Closes https://github.com/OpenZeppelin/rust-contracts-stylus/issues/542
Towards https://github.com/OpenZeppelin/rust-contracts-stylus/issues/540


packages/core/stylus/src/contract.ts
```diff
@@ -39,7 +39,7 @@ export interface BaseFunction {
   name: string;
   args: Argument[];
   code: string[];
-  visibility?: 'pub';
+  visibility?: 'pub'; // TODO: check if at all necessary
```
bidzyyys, line 42: I believe it is not necessary.

packages/ui/src/stylus/ERC1155Controls.svelte
```diff
@@ -0,0 +1,59 @@
...
+  <div class="checkbox-group">
+    <label class:checked={opts.burnable}>
+      <input type="checkbox" bind:checked={opts.burnable}>
+      Burnable
+      <HelpTooltip>
+        Token holders will be able to destroy their tokens.
+      </HelpTooltip>
+    </label>
+
+    <!-- <label class:checked={opts.pausable}>
```
bidzyyys, line 45: Add TODO info here

packages/core/stylus/src/erc1155.ts
```diff
@@ -0,0 +1,152 @@
...
+  return {
+    ...opts,
+    ...withCommonContractDefaults(opts),
+    burnable: opts.burnable ?? defaults.burnable,
+    pausable: opts.pausable ?? defaults.pausable,
+  };
+}
+
+export function isAccessControlRequired(opts: Partial<ERC1155Options>): boolean {
+  return opts.pausable === true;
```
bidzyyys, line 38: Why `opts.pausable` here? can't you just return `opts.pausable` without `===`?

bidzyyys (approved):
Looks good to me!

packages/core/stylus/src/erc1155.ts
0xNeshi: This is scaffold code, the explicit `=== true` check was needed because `opts.pausable` is _partial_, i.e. can be `undefined`.

We can actually make `opts: Required<ERC1155Options>`, making `opts.pausable` always defined, so the check can be simplified.

PR #450, opened 2025-02-20
Fix modifiers order to follow Solidity style guides

GianfrancoBazzani:
This PR fixes the function printer to respect the modifier order suggested by [Solidity Style Guide](https://docs.soliditylang.org/en/latest/style-guide.html)  

Example, before:

image

Example, after:

image


packages/core/solidity/src/print.ts
```diff
@@ -166,19 +166,22 @@ function printFunction(fn: ContractFunction, helpers: Helpers): Lines[] {
...
   }
-
+  
   if (fn.override.size === 1) {
     modifiers.push(`override`);
   } else if (fn.override.size > 1) {
     modifiers.push(`override(${[...fn.override].map(transformName).join(', ')})`);
   }
 
+  if (fn.modifiers) {
```
ericglau, line 181: `fn.modifiers` is an array so this always evalutes to `true`.  We could either remove the if-condition, or change it to check for non-empty array.

packages/core/solidity/src/zip-foundry.ts
```diff
@@ -272,15 +272,15 @@ then
...
 ${c.upgradeable ? `\
   # Install OpenZeppelin Contracts and Upgrades
-  forge install OpenZeppelin/openzeppelin-contracts-upgradeable@v${contracts.version} --no-commit --quiet
-  forge install OpenZeppelin/openzeppelin-foundry-upgrades --no-commit --quiet\
+  forge install OpenZeppelin/openzeppelin-contracts-upgradeable@v${contracts.version} --no-git --quiet
+  forge install OpenZeppelin/openzeppelin-foundry-upgrades --no-git --quiet\
 ` : `\
   # Install OpenZeppelin Contracts
-  forge install OpenZeppelin/openzeppelin-contracts@v${contracts.version} --no-commit --quiet\
+  forge install OpenZeppelin/openzeppelin-contracts@v${contracts.version} --no-git --quiet\
```
ericglau, line 283: The `--no-commit` option still seems to be available in [Forge v1](https://book.getfoundry.sh/reference/forge/forge-install#vcs-options), and I tested that it works locally.

`--no-git` changes the behavior to not use git submodules at all, but that makes it hard for users to update their dependencies later on.

We should revert this part if it still works, so that the project setup more closely aligns with Foundry conventions.

ericglau (commented):
Just a few comments, thanks! Can you also add an entry at the top of `packages/core/solidity/CHANGELOG.md` in a new `Unreleased` section?

packages/core/solidity/src/print.ts
GianfrancoBazzani: fixed

README.md
```diff
@@ -13,6 +13,8 @@ Contracts Wizard is a web application to interactively build a contract out of c
 Install dependencies with `yarn install`.
 
 `packages/core` contains the code generation logic for each language under separately named subfolders.
```
ericglau, line 15:
```suggestion
`packages/core` contains the code generation logic for each language under separately named subfolders. From each language's subfolder:
```

ericglau (approved):
LGTM, thanks!

PR #452, opened 2025-02-21
Add linter (and formatter through eslint), config and ci action step

### LINKED ISSUE #429: Add TypeScript linter
Add TypeScript linter.  This should cover .ts files in the various packages under the `packages` directory.

See linter in the [openzeppelin-upgrades repo](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/6c075654264da71050d283ea6668bef921a48df6/package.json#L9) for example.

### LINKED ISSUE #446: Add TS linter rule when import not used as value
Enable a linter rule to give an error if a value import is only used as a type.  This replaces the need for the TS compiler option [importsNotUsedAsValues](https://www.typescriptlang.org/tsconfig/#importsNotUsedAsValues) which is no longer supported in TypeScript v5 and was removed in https://github.com/OpenZeppelin/contracts-wizard/pull/231.

Depends on https://github.com/OpenZeppelin/contracts-wizard/issues/429

CoveMB:
Add eslint and required packages.
Add recommended eslint configuration (including prettier and unicorn) for various files type (similarly to what is found in openzeppelin-upgrades).
Make changes so that the current codebase complies with added lint rules.
Add lint step in CI action.
Run linter with --fix to adjust prettier rules.

Fixes #429 
Fixes #446


ericglau:

packages/core/solidity/src/contract.ts
```diff
@@ -53,15 +53,13 @@ export interface ContractFunction extends BaseFunction {
...
+export type FunctionMutability = (typeof mutabilityRank)[number];
 
 // Order is important
 const mutabilityRank = ['pure', 'view', 'nonpayable', 'payable'] as const;
 
 function maxMutability(a: FunctionMutability, b: FunctionMutability): FunctionMutability {
-  return mutabilityRank[
-    Math.max(mutabilityRank.indexOf(a), mutabilityRank.indexOf(b))
-  ]!;
+  return mutabilityRank[Math.max(mutabilityRank.indexOf(a), mutabilityRank.indexOf(b))]!;
```
ericglau, line 62: This gives a `Forbidden non-null assertion` warning, but looks like valid usage.  Other files with this warning look valid too.  I think we can remove the `'@typescript-eslint/no-non-null-assertion'` rule in the eslint config.

.github/workflows/test.yml
```diff
@@ -31,6 +31,8 @@ jobs:
       - name: Check Svelte
         run: yarn svelte-check
         working-directory: packages/ui
+      - name: Run linter
+        run: yarn lint
```
ericglau, line 35: Can we move the linter into a separate job in this file?  Since it should run on the entire repository and is not specific to a language like Solidity or Cairo.

eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
...
+  typescriptEslint.configs.strict,
+  prettierPluginRecommended,
+  {
+    plugins: {
+      unicorn: unicornPlugin,
+    },
+    rules: {
+      '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_', varsIgnorePattern: '^_' }],
+      '@typescript-eslint/no-non-null-assertion': 'warn',
+      '@typescript-eslint/consistent-type-imports': 'error',
```
ericglau, line 19: Nice, this fixes https://github.com/OpenZeppelin/contracts-wizard/issues/446

ericglau (commented):
This is looking great, thanks @CoveMB!  Can you add some content under the Development section of README.md on how to run lint and lint --fix?

.github/workflows/test.yml
```diff
@@ -6,6 +6,20 @@ on:
   pull_request: {}
 
 jobs:
+  lint:
+    name: Lint Codebase
```
ericglau, line 10: We could remove the name so it just uses the id `lint`, which is clear enough.
```suggestion
```

ericglau (approved):
LGTM, thanks!

PR #454, opened 2025-02-21
Run command utils

CoveMB:
A small script util to run some commands directly in root folder


scripts/run-command.mjs
```diff
@@ -0,0 +1,5 @@
+import { commandAndFlagsInput } from "./command-input.mjs";
+import { inLanguageFolderRunTYarnCommand } from "./execute-command.mjs";
+import { languageInput } from "./language-input.mjs";
+
+inLanguageFolderRunTYarnCommand(languageInput, commandAndFlagsInput);
```
ericglau, line 5: It seems odd to pass through constants which are evaluated by some other functions that are not visible here.

Could the constants be moved to this file instead?  For example, this looks more readable:
```
const languageInput = getAndCheckLanguageInput();
const commandAndFlagsInput = getCommandAndFlagsInput().join(" ");

runYarnCommandForLanguage(languageInput, commandAndFlagsInput);
```

ericglau (approved):

scripts/run-command.mjs
CoveMB: Good point

scripts/command-input.mjs
```diff
@@ -0,0 +1,8 @@
+export function getCommandAndFlagsInput() {
+  const [, , _languageFolderName, ...commandAndFlags] = process.argv;
+
+  if (commandAndFlags) return commandAndFlags.join(" ");
+
+  console.error("Please provide a command.");
```
ericglau, line 6:
```suggestion
  console.error("Provide a command.");
```

scripts/language-input.mjs
```diff
@@ -0,0 +1,34 @@
...
+    return statSync(subfolderPath).isDirectory();
+  });
+}
+
+function getLanguageInput() {
+  const [, , languageFolderName] = process.argv;
+
+  if (languageFolderName) return languageFolderName;
+
+  console.error("Please provide a language name.");
```
ericglau, line 18:
```suggestion
  console.error("Provide a language name.");
```

Issue #212, opened 2023-01-05
ERC721 safeMint should return the tokenId


nkavian:
When auto increment ids are enabled, shouldn't the method also then `return tokenId`? Otherwise how would you know what token was minted?

<img width="798" alt="image" src="https://user-images.githubusercontent.com/1127421/210900621-528a4449-f7dd-4bec-b64e-6c9e11e37764.png">

frangio:
> Otherwise how would you know what token was minted?

Do you mean when you invoke this function off-chain? In that case you would use the `Transfer` event, and in fact the return value wouldn't be available.

Or do you have another scenario in mind?

nkavian:
The contract caller doesn't know the value of the new tokenId that was generated. I'm suggesting an implementation like this:

```
function safeMint(address to) public onlyOwner returns (uint256) {
    uint256 tokenId = _tokenIdCounter.current();
    _tokenIdCounter.increment();
    _safeMint(to, tokenId);
    return tokenId;
}
```

EthixLucifer:
can i work on this issue @ericglau

PR #455, opened 2025-02-21
Add return tokenId (and adjust type) if incremental safeMint

CoveMB:
Add return tokenId and  and adjust to return type to uint256 if incremental safeMint, if incremental safeMint

Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/212


ericglau (approved):
LGTM, thanks! Can you add an entry to the changelog in `packages/core/solidity/CHANGELOG.md`?

PR #460, opened 2025-02-25
Add Stellar Contracts

brozorec:
This PR adds support for the available modules in Stellar Contracts Library [v0.1.0](https://github.com/OpenZeppelin/stellar-contracts/releases/tag/v0.1.0)


ericglau:

ericglau:
@SocketSecurity ignore npm/undici@5.28.4

Unrelated to this PR, see https://github.com/OpenZeppelin/contracts-wizard/pull/430#issuecomment-2602776853

packages/core/stellar/src/fungible.ts
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    // TODO: handle signed int?
+    const premintAbsolute = toUint(getInitialSupply(amount, 18), 'premint', 'u128');
+
+    c.addUseClause('openzeppelin_fungible_token', 'mintable::FungibleMintable');
+    c.addUseClause('soroban_sdk', 'Address');
+
+    c.addConstructorArgument({ name: 'owner', type: 'Address' });
+    c.addConstructorCode(`fungible::mintable::mint(e, &owner, ${premintAbsolute});`);
```
ericglau, line 156: A recommended pattern in our other contracts libraries are to use a separate constructor argument for the recipient of premints, rather than preminting to the owner.

For example, this is done for both the Solidity and Cairo Wizards, where premint goes to a `recipient` which is a separate constructor parameter from `owner`.

What do you think of doing that here as well?

ericglau (approved):
Just a couple of questions, thanks

packages/core/stellar/src/fungible.ts
brozorec: Seems fine 👍 
modified with https://github.com/OpenZeppelin/contracts-wizard/pull/460/commits/ca64c58379d61d20a294cefb803bf33e505eab90

packages/core/stellar/src/utils/convert-strings.ts
```diff
@@ -0,0 +1,83 @@
+import { OptionsError } from '../error';
+
+// TODO: Review and convert these to properly handle language restrictions for Stellar
```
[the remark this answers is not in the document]
brozorec: Yes, this looks fine for the moment 👍

PR #469, opened 2025-03-03
chore: fix language input validation and improve error handling

hexcow:
I’ve fixed an issue with the language input validation. Now, the `validateLanguageInput` function checks that the input only contains valid characters (letters, numbers, dashes, and underscores) using a regular expression. This ensures that only correct and safe input is accepted, preventing potential vulnerabilities.  

Additionally, I improved the error handling. The catch block now logs a more detailed error message: if an error message is available, it will be displayed first, followed by the error stack or the error object itself if no message is present.

### DIFF
--- a/scripts/execute-command.mjs
+++ b/scripts/execute-command.mjs
@@ -1,12 +1,20 @@
 import { execSync } from 'child_process';
 
+function validateLanguageInput(languageInput) {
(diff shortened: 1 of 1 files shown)


CoveMB:
Hi @hexcow thanks for submitting this fix but we don't think it is necessary at the moment has this only impact a development related script, closing for now.

PR #470, opened 2025-03-03
Add Stylus Contracts

0xNeshi:
Resolves https://github.com/OpenZeppelin/rust-contracts-stylus/issues/568
Towards https://github.com/OpenZeppelin/rust-contracts-stylus/issues/584


packages/ui/public/stylus.html
```diff
@@ -0,0 +1,82 @@
...
+  </div>
+  <div class="wizard-container">
+
+    <oz-wizard class="wizard" data-sync-url="fragment" data-lang="stylus"></oz-wizard>
+  </div>
+
+  <footer>
+    <p>© <a href="https://openzeppelin.com" target="_blank" rel="noopener noreferrer">OpenZeppelin</a> 2017-2025 |&nbsp;<a href="https://openzeppelin.com/privacy" target="_blank" rel="noopener noreferrer">Privacy</a> |&nbsp;<a href="https://openzeppelin.com/tos" target="_blank" rel="noopener noreferrer">Terms of Service</a></p>
+  </footer>
+
```
0xNeshi, line 66: @ericglau what do you think about calculating the current copyright year automatically?
```suggestion
  <footer>
    <p>© <a href="https://openzeppelin.com" target="_blank" rel="noopener noreferrer">OpenZeppelin</a> 2017-<span id="currentYear"></span> |&nbsp;<a href="https://openzeppelin.com/privacy" target="_blank" rel="noopener noreferrer">Privacy</a> |&nbsp;<a href="https://openzeppelin.com/tos" target="_blank" rel="noopener noreferrer">Terms of Service</a></p>
  </footer>
```
and then we could add a script below:
```javascript
<script>
    document.getElementById('currentYear').textContent = new Date().getFullYear();
</script>
```

ericglau: Makes sense. A similar approach is used for the docs [here](https://github.com/OpenZeppelin/docs.openzeppelin.com/blob/master/ui/theme/partials/footer.hbs#L3) and calculated [here](https://github.com/OpenZeppelin/docs.openzeppelin.com/blob/master/ui/src/js/copyright-date.js).

Could you make the change for the other langs as well (index.html, cairo.html, stellar.html)?

packages/core/stylus/src/erc721.ts
```diff
@@ -0,0 +1,218 @@
...
+  if (allOpts.pausable) {
+    addPausable(c, allOpts.access);
+  }
+
+  if (allOpts.burnable) {
+    addBurnable(c, allOpts.pausable, allOpts.enumerable);
+  }
+
+  if (allOpts.enumerable) {
+    addEnumerable(c, allOpts.pausable);
```
ericglau, line 63: Should Enumerable be in the UI or is it not ready yet?

packages/core/stylus/src/erc20.ts
```diff
@@ -0,0 +1,262 @@
...
+import { withCommonContractDefaults, getSelfArg } from './common-options';
+import { contractDefaults as commonDefaults } from './common-options';
+import { printContract } from './print';
+import { setAccessControl } from './set-access-control';
+import { setInfo } from './set-info';
+
+export interface ERC20Options extends CommonContractOptions {
+  name: string;
+  burnable?: boolean;
+  pausable?: boolean;
```
ericglau, line 15: Pausable is hidden from the UI but it could still be accessible via the API (for all of the contract kinds).  Options to consider:
- Comment out pausable logic from each contract kind, or
- Exclude pausable from types in `api.ts` (although users could still access it by ignoring type safety), or
- Do not release the NPM package with the API until the project is more developed

packages/core/stylus/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@openzeppelin/wizard-stylus",
+  "version": "0.1.0",
```
ericglau, line 3: Do we intend to target a full or alpha version of the Contracts for Stylus library?  If alpha, we could consider making this package version alpha as well.

packages/ui/src/stylus/ERC20Controls.svelte
```diff
@@ -0,0 +1,76 @@
...
+      Permit
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts-stylus/erc20-permit">
+        Without paying gas, token holders will be able to allow third parties to transfer from their account.
+      </HelpTooltip>
+    </label>
+
+    <label class:checked={opts.flashmint}>
+      <input type="checkbox" bind:checked={opts.flashmint}>
+      Flash Minting
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts-stylus/erc20-flash-mint">
```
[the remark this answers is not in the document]
0xNeshi: I'll check this and get back to you. In case alpha versions won't be published, is it alright if we just omit these links until the official 0.2.0 version is published?

packages/ui/src/stylus/ERC1155Controls.svelte
```diff
@@ -0,0 +1,68 @@
...
+      Burnable
+      <HelpTooltip>
+        Token holders will be able to destroy their tokens.
+      </HelpTooltip>
+    </label>
+
+    <label class:checked={opts.supply}>
+      <input type="checkbox" bind:checked={opts.supply}>
+      Supply Tracking
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts-stylus/erc1155-supply">
```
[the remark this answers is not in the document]
0xNeshi: See https://github.com/OpenZeppelin/contracts-wizard/pull/470#discussion_r1979330423

packages/core/stylus/src/erc721.ts
0xNeshi: oops, nice catch!

packages/core/stylus/src/erc20.ts
0xNeshi: Commented out the logic

packages/core/stylus/src/print.ts
```diff
@@ -0,0 +1,290 @@
...
+  if (!contract.eip712Needed) {
+    return [];
+  }
+
+  return [
+    '#[storage]',
+    'struct Eip712 {}',
+    '',
+    'impl IEip712 for Eip712 {',
+    spaceBetween([`const NAME: &'static str = "${contract.name}";`, `const VERSION: &'static str = "1";`]),
```
ericglau, line 180: Note this uses the sanitized name from the identifier, which may be overly restrictive for strings.

When we add support for constructors, the string token name format should be less restrictive, and this EIP712 name should ideally match the string token name.

For reference, Cairo sanitizes strings (depending on scenario) with [these two functions](https://github.com/OpenZeppelin/contracts-wizard/blob/e284fe4650f05cb4f9dd6385bc1395fb165e4149/packages/core/cairo/src/utils/convert-strings.ts#L23-L51).

packages/core/stylus/package.json
ericglau: Note the version of this package won't be guaranteed to always match with the version of the library.  But I think it makes sense to start with the same version.

packages/core/stylus/README.md
```diff
@@ -0,0 +1,86 @@
...
+function isAccessControlRequired(opts: Partial<ERC20Options>): boolean
+```
+```js
+function isAccessControlRequired(opts: Partial<ERC721Options>): boolean
+```
+```js
+function isAccessControlRequired(opts: Partial<ERC1155Options>): boolean
+```
+Whether any of the provided options require access control to be enabled. If this returns `true`, then calling `print` with the same options would cause the `access` option to default to `'ownable'` if it was `undefined` or `false`. 
+
```
ericglau, line 57: Access control options are also exposed through the API and currently have no effect.  Instead of commenting out the code and introducing unnecessary churn, I think we can just document it as such:
```suggestion

> [!NOTE]
> Access control options are under development and setting `access` currently has no effect.

```

ericglau (approved):
Just a few comments as above.  The rest LGTM, thanks!

packages/core/stylus/src/print.ts
0xNeshi: You're right. I [updated the code](https://github.com/OpenZeppelin/contracts-wizard/pull/470/commits/c75c0e7048f188a753920bdff46f794dcb96570d) to allow unicode string literals for token name and this `const NAME`.

Note that I added a new function `escapeString` which escapes only double quotes and backslashes to align with how other similar string sanitizers behave (e.g. the Cairo ones you linked to). Lmk if this is ok or if you'd like to handle other special characters.

ericglau (approved):

packages/core/stylus/src/print.ts
ericglau: Looks good, thanks.

PR #471, opened 2025-03-03
Fix missing `no_std` in Stellar contracts

brozorec:
Fix https://github.com/OpenZeppelin/stellar-contracts/issues/91 by adding `#![no_std]` to all generated output.


packages/core/stellar/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.1 (2025-03-03)
+
+- Add a default no_std to all contracts. (Fix [#91](https://github.com/OpenZeppelin/stellar-contracts/issues/91))
```
ericglau, line 5: We should use the Wizard PR here number for consistency.
```suggestion
- Add a default no_std to all contracts. ([#471](https://github.com/OpenZeppelin/contracts-wizard/pull/471))
```

brozorec: noted 👍

PR #476, opened 2025-03-05
Fix: Automatically use unicode syntax for token names with non-ASCII characters

### LINKED ISSUE #475: Solidity: If a dev passes a unicode string as token name, `unicode"..."` string literal should be used
Currently passing Unicode strings as token names is supported in Wizard, but it produces invalid Solidity code. 

For example, if we open Wizard Solidity and in the "Name" field we type `MyTokeć`, this will be the output:

```solidity
// Compatible with OpenZeppelin Contracts ^5.0.0
pragma solidity ^0.8.22;

import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
import {ERC20Permit} from "@openzeppelin/contracts/token/ERC20/extensions/ERC20Permit.sol";

contract MyTokec is ERC20, ERC20Permit {
    constructor() ERC20("MyTokeć", "MTK") ERC20Permit("MyTokeć") {}
}
```

But this is not valid Solidity code as [string literals must be ASCII characters](https://docs.soliditylang.org/en/latest/types.html#string-literals-and-types), so the compiler will throw an error. To make this compile, the [string literal must be prepended with the `unicode` keyword](https://docs.soliditylang.org/en/latest/types.html#unicode-literals):

```solidity
contract MyTokec is ERC20, ERC20Permit {
    constructor() ERC20(unicode"MyTokeć", "MTK") ERC20Permit(unicode"MyTokeć") {}
}
```

MKVEERENDRA:
Fix: Use unicode string literal for token names with non-ASCII characters (#475)

- Token names with accents/symbols (e.g., "MyTokeć") now use Solidity’s `unicode"..."` syntax   to prevent compilation errors.
- Double quotes in names/symbols are escaped (`\"`) to avoid syntax issues.
- Developers can now use any Unicode name without manual code adjustments.

Fixes #475


packages/core/solidity/src/erc20.ts
```diff
@@ -109,13 +109,22 @@ export function buildERC20(opts: ERC20Options): ContractBuilder {
 
   return c;
 }
+// Helper function to format string literals with the `unicode` keyword if they contain non-ASCII characters.
+// Also escapes double quotes in the string.
+function formatLiteral(str: string): string {
```
CoveMB, line 114: Let's try to make the function name clear and specific enough so we might not need the comment

CoveMB, line 114: We could move this in the utils folder

packages/core/solidity/src/erc20.ts
```diff
@@ -109,13 +109,22 @@ export function buildERC20(opts: ERC20Options): ContractBuilder {
...
+}
 
 function addBase(c: ContractBuilder, name: string, symbol: string) {
   const ERC20 = {
     name: 'ERC20',
     path: '@openzeppelin/contracts/token/ERC20/ERC20.sol',
   };
-  c.addParent(ERC20, [name, symbol]);
+  // Use formatLiteral to wrap name and symbol appropriately
+  c.addParent(ERC20, [formatLiteral(name), formatLiteral(symbol)]);
```
CoveMB, line 127: great, we should apply this to other contract (not only ERC20) as well maybe in the contract builder

CoveMB:
Hey @MKVEERENDRA thanks for submitting this fix ⚡
left some comments, also we might want to add some test

CoveMB:
Hey @MKVEERENDRA please consider signing the CLA

MKVEERENDRA:
@ericglau any suggestions

CoveMB:
Hi, @MKVEERENDRA you'll need to sign the CLA

MKVEERENDRA:

MKVEERENDRA:
recheck

ericglau (approved):
LGTM, thanks!

I noticed Open in Remix causes an error if the contract has Unicode, but opened a [separate issue here](https://github.com/OpenZeppelin/contracts-wizard/issues/545).

CoveMB (approved):

PR #478, opened 2025-03-06
Stylus updates for v0.2.0-alpha.4 + Fix Broken Templates

0xNeshi:
Resolves https://github.com/OpenZeppelin/rust-contracts-stylus/issues/593

We released a new Stylus version https://github.com/OpenZeppelin/rust-contracts-stylus/releases/tag/v0.2.0-alpha.4

Updated the templates to make them compatible. Also fixed some broken templates.


packages/core/stylus/src/erc1155.ts
```diff
@@ -93,11 +95,14 @@ function addBase(c: ContractBuilder, allOpts: ERC1155Options): BaseImplementedTr
...
   // c.addFunctionCodeBefore(baseTrait, functions(baseTrait).safe_batch_transfer_from, ['self.pausable.when_not_paused()?;']);
   // }
-
-  return baseTrait;
 }
 
-function addSupplyFunctions(c: ContractBuilder) {
+// This adds supply-related parts without re-adding the contract structure,
+// as it was already added in `addBase`.
+function addSupply(c: ContractBuilder) {
```
ericglau, line 102: The caller could be using `erc1155SupplyTrait` or `erc1155Trait`.  Meanwhile, this function is using `erc1155SupplyTrait`.

While the caller only calls this function if `allOpts.supply` is true, it may be safer to add an assertion that this function is indeed called while the caller has `erc1155SupplyTrait`.

ericglau (approved):
LGTM, just one minor suggestion.

qalisander (approved):
LGTM

packages/core/stylus/src/erc1155.ts
0xNeshi: Supply-related functions are always called on `erc1155SupplyTrait`. Nevertheless, having an assertion is always better, will add it.

bidzyyys (commented):
Looks good!

PR #481, opened 2025-03-11
Support Cairo alpha version

immrsd:
Adds support for Cairo alpha version by:
- duplicating `core/cairo` as `core/cairo-alpha`
- duplicating `ui/cairo` as `ui/cairo-alpha`
- adding a switch control to switch between the audited (stable) version and the latest alpha

image

image

image


ericglau:
Thanks @immrsd!

Added the following:
- Dynamically updates the URL params when changing the version and contract tab, so that the same page can be accessed by copy/pasting the URL.  Example: [/cairo#erc1155&version=2.0.0-alpha](https://deploy-preview-481--openzeppelin-contracts-wizard.netlify.app/cairo#erc1155&version=2.0.0-alpha)
- For the embed, the version switcher will be visible if no version is specified in the URL.  But if a version is specified in the URL, it will just show the requested version and not show the switcher.  This is intended to allow embedding on version-specific docs sites.
  - Example: Embed without version param (shows switcher): [/embed?lang=cairo](https://deploy-preview-481--openzeppelin-contracts-wizard.netlify.app/embed?lang=cairo)
  - Example: Embed with version param (hides switcher): [/embed?lang=cairo&version=2.0.0-alpha](https://deploy-preview-481--openzeppelin-contracts-wizard.netlify.app/embed?lang=cairo&version=2.0.0-alpha)
  - If for some reason we want to show the switcher even if there is a version param, we can include `force-version-switch=true`.
    - Example: [/embed?lang=cairo&version=2.0.0-alpha&force-version-switch=true](https://deploy-preview-481--openzeppelin-contracts-wizard.netlify.app/embed?lang=cairo&version=2.0.0-alpha&force-version-switch=true)
    - This is also provided by the main Cairo Wizard webpage via `cairo.html`.
- Evaluates app selection based on semver matching.
  - Example: Requesting version 2.0.0 which is compatible with the alpha: [/embed?lang=cairo&version=2.0.0](https://deploy-preview-481--openzeppelin-contracts-wizard.netlify.app/embed?lang=cairo&version=2.0.0)
  - Example: Requesting version 1.0.0 which is compatible with stable: [/embed?lang=cairo&version=1.0.0](https://deploy-preview-481--openzeppelin-contracts-wizard.netlify.app/embed?lang=cairo&version=1.0.0)
  - Example: Requesting version 3.0.0 which not compatible with any supported range: [/embed?lang=cairo&version=3.0.0](https://deploy-preview-481--openzeppelin-contracts-wizard.netlify.app/embed?lang=cairo&version=3.0.0)

Questions:
- Should we support tag aliases for the URL as an alternative to specific version numbers, so that the URLs can be more long-lived and always point to the most relevant version?
  - Example: Do we want to support something like `/cairo#erc1155&version=stable` and `/cairo#erc1155&version=alpha` ?

packages/core/cairo_alpha/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@openzeppelin/wizard-cairo-alpha",
+  "version": "1.0.0",
```
ericglau, line 3: We probably don't want to publish these packages, especially since they are using a different package name, so we could mark this as private.  If we do want to publish an alpha package, we may want to do that from a branch and handle it case by case (probably with `@openzeppelin/wizard-cairo` with an alpha version number).

```suggestion
  "private": true,
  "version": "1.0.0",
```

packages/ui/src/cairo/App.svelte
```diff
@@ -35,6 +35,10 @@
       dispatch('tab-change', tab);
     };
 
+    const NAV_BAR_OFFSET = 84;
+    const VERSION_SWITCH_OFFSET = 40;
+    const dynamicHeight = `calc(100vh-${NAV_BAR_OFFSET + VERSION_SWITCH_OFFSET}px)`;
```
ericglau, line 40: This doesn't seem to work well, for example on Vesting or Custom tabs where there are very few options, the height is very small.  I think we should revert this, and attempt to rework the height calculation for all of the languages in a separate PR.

Same for `cairo_alpha/App.svelte`

ericnordelo (approved):
LGTM!

ericglau (approved):
LGTM. Can we revert the `package-lock.json` changes?  That has a lot of changes and shouldn't be needed for this PR.

ericnordelo:
Let's wait until the alpha get's released before merging this. And we should target this branch for the alpha wizard update @immrsd. So that get's merged to main at the same time after the release.

ericglau:
Merged updates from master's cairo UI into cairo_alpha UI

ericglau:
@SocketSecurity ignore-all
Unrelated to this PR.

PR #486, opened 2025-03-15
Add Account and increase pragma to 0.8.27

ernestognw:
### Description

Adds support for Accounts in OpenZeppelin Community Contracts.


ernestognw:
Seems like the build is broken because some libraries of the community repository rely on the `Calldata` library, which is not released yet until 5.3

ernestognw:
I keep getting the following error when running `yarn prepare`:

```
Downloading compiler 0.8.27
 | TypeError: Linearization of inheritance graph impossible
 |   --> contracts/generated/504cdf2b09d152566e5b55ac062f0ff8e65fc29c.sol:14:1:
 |    |
 | 14 | contract MyAccount is Initializable, Account, EIP712, ERC7739, SignerRSA, ERC7821, ERC721Holder, ERC1155Holder {
 |    | ^ (Relevant source part starts here and spans across multiple lines).
 | 
 | 
 | TypeError: Linearization of inheritance graph impossible
 |   --> contracts/generated/0a34ddf6c6eac317acfc9d6a7ad3badf6c2aa72f.sol:16:1:
 |    |
 | 16 | contract MyAccount is Initializable, Account, EIP712, AccountERC7579, ERC7739, SignerECDSA, ERC721Holder, ERC1155Holder {
 |    | ^ (Relevant source part starts here and spans across multiple lines).
```

However, the output seems fine and is compiling independently. Here's an example:

```solidity
// Compatible with OpenZeppelin Contracts ^5.0.0
pragma solidity ^0.8.27;

import {AbstractSigner} from "@openzeppelin/community-contracts/utils/cryptography/AbstractSigner.sol";
import {Account} from "@openzeppelin/community-contracts/account/Account.sol";
import {AccountERC7579} from "@openzeppelin/community-contracts/account/extensions/AccountERC7579.sol";
import {EIP712} from "@openzeppelin/contracts/utils/cryptography/EIP712.sol";
import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
import {ERC721Holder} from "@openzeppelin/contracts/token/ERC721/utils/ERC721Holder.sol";
import {ERC7739} from "@openzeppelin/community-contracts/utils/cryptography/ERC7739.sol";
import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
import {PackedUserOperation} from "@openzeppelin/contracts/interfaces/draft-IERC4337.sol";
import {SignerECDSA} from "@openzeppelin/community-contracts/utils/cryptography/SignerECDSA.sol";

contract MyAccount is Initializable, Account, EIP712, AccountERC7579, ERC7739, SignerECDSA, ERC721Holder, ERC1155Holder {
    constructor() EIP712("MyAccount", "1") {}

    function isValidSignature(bytes32 hash, bytes calldata signature)
        public
        view
        override(AccountERC7579, ERC7739)
        returns (bytes4)
    {
        // ERC-7739 can return the fn selector (success), 0xffffffff (invalid) or 0x77390001 (detection).
        // If the return is 0xffffffff, we fallback to validation using ERC-7579 modules.
        bytes4 erc7739magic = ERC7739.isValidSignature(hash, signature);
        return erc7739magic == bytes4(0xffffffff) ? AccountERC7579.isValidSignature(hash, signature) : erc7739magic;
    }

    function _rawSignatureValidation(bytes32 hash, bytes calldata signature)
        internal
        view
        override(SignerECDSA, AbstractSigner, AccountERC7579)
        returns (bool)
    {
        // Force signer validation first, and fallback to ERC-7579
        // return SignerECDSA._rawSignatureValidation(hash, signature) || AccountERC7579._rawSignatureValidation(hash, signature);
        return super._rawSignatureValidation(hash, signature);
    }

    function initializeECDSA(address signer) public initializer {
        _setSigner(signer);
    }

    // The following functions are overrides required by Solidity.

    function _validateUserOp(PackedUserOperation calldata userOp, bytes32 userOpHash)
        internal
        override(Account, AccountERC7579)
        returns (uint256)
    {
        return super._validateUserOp(userOp, userOpHash);
    }
}
```

CoveMB:
The `linearization` issue might be due to the inheriting from many contracts, some might share common ancestor? Maybe you could check if that is the case with `solc --ast-compact-json`

ernestognw:
> The `linearization` issue might be due to the inheriting from many contracts, some might share common ancestor? Maybe you could check if that is the case with `solc --ast-compact-json`

I'm familiar with the linearization issue, but the current implementation only produces contracts that can be linearized using Solidity's C3 algorithm for resolution. However, I think it's a dependency issue, but upgrading `@openzeppelin/community-contracts` fails since the newest version expects things from OZ Contracts in master (i.e. ENTRYPOINT_V08).

Maybe we'll need to wait until we release 5.4 😢

CoveMB:
Ha gotcha, maybe you can check that with `yarn add @openzeppelin/contracts@github:OpenZeppelin/openzeppelin-contracts` to check if it works with (unreleased) code from master

ernestognw:
> Ha gotcha, maybe you can check that with `yarn add @openzeppelin/contracts@github:OpenZeppelin/openzeppelin-contracts` to check if it works with (unreleased) code from master

Great idea! Though it seemed to pull the latest release tag so also failed. We just released 5.3 so now it should be fine, will keep working on this. Thanks @CoveMB

packages/core/solidity/src/api.ts
```diff
@@ -59,6 +65,7 @@ export type ERC721 = WizardContractAPI<ERC721Options>;
 export type ERC1155 = WizardContractAPI<ERC1155Options>;
 export type Stablecoin = WizardContractAPI<StablecoinOptions>;
 export type RealWorldAsset = WizardContractAPI<StablecoinOptions>;
+export type Account = WizardContractAPI<AccountOptions>;
```
ericglau, line 68: nit: since access control is not relevant for Account, we could remove `isAccessControlRequired` from the Account API by doing something similar to [Cairo](https://github.com/OpenZeppelin/contracts-wizard/blob/18a4baf51d4004ec209d7c38bcbe3fdd77cca437/packages/core/cairo/src/api.ts#L35-L74) by defining different interfaces for Account vs regular contracts, or perhaps by simply omitting the field from the type here.

packages/ui/src/solidity/App.svelte
```diff
@@ -83,6 +84,7 @@
           case 'ERC1155':
           case 'Stablecoin':
           case 'RealWorldAsset':
+          case 'Account':
```
ericglau, line 87: On line 151, can you also add 'Account' to the conditions such that:
```
    } else if (
      opts?.kind === 'Stablecoin' ||
      opts?.kind === 'RealWorldAsset' ||
      opts?.kind === 'Account' ||
      (opts?.kind === 'ERC20' && opts.crossChainBridging)
    ) {
      return {
        openInRemix: false,
        downloadHardhat: false,
        downloadFoundry: false,
      };
    }
```
Community Contracts doesn't work with current Remix or Hardhat package options for now, since there is no npm release.  And the Foundry package isn't set up for Community Contracts yet.

packages/core/solidity/src/generate/account.ts
```diff
@@ -0,0 +1,22 @@
...
+const account = {
+  name: ['MyAccount'],
+  ERC1271: [false, 'ERC1271', 'ERC7739'] as const,
+  ERC721Holder: [false, true] as const,
+  ERC1155Holder: [false, true] as const,
+  signer: ['ERC7702', 'ECDSA', 'P256', 'RSA'] as const,
+  ERC7821: [false, true] as const,
+  ERC7579: [false, 'AccountERC7579', 'AccountERC7579Hooked'] as const,
+  access: accessOptions,
+  upgradeable: upgradeableOptions,
```
ericglau, line 16: We should be able to remove `access` and `upgradeable` from here

packages/core/solidity/hardhat.config.js
```diff
@@ -41,5 +41,14 @@ task(TASK_COMPILE_SOLIDITY_MERGE_COMPILATION_JOBS, async ({ compilationJobs }, _
  * @type import('hardhat/config').HardhatUserConfig
  */
 module.exports = {
-  solidity: SOLIDITY_VERSION,
+  solidity: {
+    version: SOLIDITY_VERSION,
+    settings: {
+      evmVersion: 'cancun',
```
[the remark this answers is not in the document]
ernestognw: Yes. The default is still `paris` in Hardhat and there seem to be [little updates](https://github.com/NomicFoundation/hardhat/issues/4851#issuecomment-1986148049) about it.

packages/core/solidity/src/generate/account.ts
ernestognw: Not really since it's defined in `CommonOptions`. I'm doing the same as [`generate/stablecolin.ts`](https://github.com/OpenZeppelin/contracts-wizard/blob/76b721fe6290dd7a500266a7d022cd8df8472e63/packages/core/solidity/src/generate/stablecoin.ts) and I'm declaring these fields as `[false] as const`

packages/core/solidity/src/api.ts
ernestognw: Amazing, I like the idea of an `AccessControlAPI` as in the Cairo definition. Updated

packages/ui/src/solidity/AccountControls.svelte
```diff
@@ -0,0 +1,187 @@
...
+      <input
+        type="checkbox"
+        checked={!!opts.ERC1271 || !!opts.ERC7579}
+        disabled={!!opts.ERC7579}
+        on:change={e => {
+          if (e.currentTarget?.checked) opts.ERC1271 = 'ERC7739';
+          else opts.ERC1271 = false;
+        }}
+      />
+      ERC-1271 Signatures
```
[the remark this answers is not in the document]
ernestognw: Yeah makes a lot of sense. I updated the names and expanded on the descriptions

packages/ui/src/solidity/AccountControls.svelte
```diff
@@ -0,0 +1,187 @@
...
+          if (e.currentTarget?.checked) opts.ERC7579 = 'AccountERC7579Hooked';
+          else opts.ERC7579 = 'AccountERC7579';
+        }}
+      />
+      Hooks
+      <HelpTooltip link="https://eips.ethereum.org/EIPS/eip-7579#hooks"
+        >Hooks enable support for pre and post checks after execution.</HelpTooltip
+      >
+    </label>
+    <label class:checked={opts.signer}>
```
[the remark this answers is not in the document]
ernestognw: Amazing suggestion, I just added `ExpandableToggleRadio` for the signer, it fits better imo

ericglau (approved):
LGTM, thanks!  I adjusted some styling and types, please review.

`upgradeable` being forced to false is fine, since we eventually want to support it.  `access` being forced to false doesn't seem very correct, since ideally we'd want to avoid having it in the Account options type in the first place.  But given this is considered experimental, I think this works for now.

ernestognw:
> LGTM, thanks! I adjusted some styling and types, please review.

LGTM too, thanks @ericglau

> upgradeable being forced to false is fine, since we eventually want to support it. 

Yes, part of future OZ vanilla releases include these contracts

> access being forced to false doesn't seem very correct, since ideally we'd want to avoid having it in the Account options type in the first place. (I tried removing the unused options but it ended up being too large of a change). But given this is considered experimental, I think this works for now.

Yeah I also tried and ended up being too much. The account has such a unique authorization model in comparison to other contracts that it might require some rework, but I agree it's fine given it's still in community contracts

PR #487, opened 2025-03-17
Update UI button labels

ericglau:
- Changes "Deploy with Defender" button to "Deploy", since some users may not be familiar with what Defender is.
  - Adds arrows to suggest that this button opens a modal rather than deploys immediately
  - Depends on changing the "Get API Key" label in the modal to "Get API Key from OpenZeppelin Defender" to provide more context. https://github.com/OpenZeppelin/defender-deploy-plugin/pull/52
- Adds "Open in Remix" text to the Open in Remix button, since the icon may not be well known
- Adds "AI Assistant" text to the AI Assistant button, since the icon may not be well known
  - Resized the button to fit text
- Adjusts layer order so the Deploy modal covers the AI Assistant, since AI Assistant does not help with Deploy

Screenshot 2025-03-18 at 11 09 27 AM

Screenshot 2025-03-17 at 8 04 51 PM


CoveMB (approved):
Looks clean

makiopen (approved):
LGTM

Issue #241, opened 2023-04-10
ERC20 Premint field has no upper bound


mark1yo:
Hi 😃 
There is a potential bug in generating an ERC20 token. The "Premint" field has no upper bound, which allows you to mint 999...9 token.

When you try to compile the generate contract, for example in Remix, you get "Literal too large" error.

hashpalk:
hey @frangio, can you assign me this issue? I want to work on this

ericglau:
@hashpalk Feel free to open a PR, thanks.

ericglau:
Fixed in https://github.com/OpenZeppelin/contracts-wizard/pull/488

PR #488, opened 2025-03-19
fix: Add validation for ERC20 premint amount to prevent "Literal too large" error

hashpalk:
This solves issue  #241 where the ERC20 token generator previously allowed arbitrarily large numbers in the premint field, which could lead to a "Literal too large" compilation error in Solidity. This occurred because very large numbers exceeded Solidity's number literal limit when combined with the token's 18 decimal places.

Changes:
- Add validation to check if the final token amount (including 18 decimals) exceeds Solidity's 77-digit literal limit
- Pre-calculate the full amount with decimals instead of using runtime multiplication
- Add clear error message when amount is too large
- Generate simpler and more direct minting code

This prevents contract compilation failures while still allowing reasonably large premint amounts up to Solidity's literal limits.


hashpalk:

packages/core/solidity/src/erc20.ts
```diff
@@ -179,11 +179,21 @@ function addPremint(
       const decimalPlace = decimals.length - exponent;
       const zeroes = new Array(Math.max(0, -decimalPlace)).fill('0').join('');
       const units = integer + decimals + zeroes;
-      const exp = decimalPlace <= 0 ? 'decimals()' : `(decimals() - ${decimalPlace})`;
+      
+      // Add 18 decimals to the number (standard ERC20 decimals)
+      const fullAmount = units + '000000000000000000';
```
ericglau, line 184: This number should not be hardcoded as 18 decimals.  While 18 is the default, users who take the generated contract as a starting point could override the [`decimals()` function of ERC20.sol](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20#ERC20-decimals--) to change the intended display decimals.  The premint amount should take this value into consideration.

For example, if the user enters a Premint of 1000 in the UI, and they override `decimals()` to 2, then the actual premint calculation should result in `100000`.

packages/core/solidity/src/erc20.ts
```diff
@@ -179,11 +179,21 @@ function addPremint(
...
+      
+      // Check if the final number would be too large
+      if (fullAmount.length > 77) { // Solidity has a limit on number literals
```
ericglau, line 187: Solidity's limit on literals is not based on character length.  Rather, this must be less than `2**256`.  There is some [documentation here](https://docs.soliditylang.org/en/latest/yul.html#literals) which is for Yul, but it seems to be the same limit for Solidity.

[Here is an example](https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/core/cairo/src/utils/convert-strings.ts#L78) of a way to validate the size of a uint (that example is currently for Cairo but something similar could be done for Solidity).

packages/core/solidity/src/erc20.ts
```diff
@@ -179,11 +179,21 @@ function addPremint(
...
+        throw new OptionsError({
+          premint: 'Amount too large. Please use a smaller number to avoid Solidity literal limits',
+        });
+      }
 
       c.addConstructorArgument({ type: 'address', name: 'recipient' });
 
-      const mintLine = `_mint(recipient, ${units} * 10 ** ${exp});`;
+      // Generate the mint line with the full amount directly
+      const mintLine = `_mint(recipient, ${fullAmount});`;
```
ericglau, line 196: The previous format should be used, so that `decimals()` from the user's contract is included in the calculation.

Overall, it seems that two checks should be made:
1. Check that the value of `units` is less than 2**256
2. Check that the calculated value of `${units} * 10 ** ${exp}` is less than 2**256, because `2**256 - 1` is the largest possible value of the `uint256` type as an argument for the `_mint` function.
    - While `decimals()` could be overridden by the user and therefore `${exp}` cannot be definitely known, for the purposes of this check we can assume `decimals()` is 18.

ericglau (commented):
Thanks for the PR.  I think the check here should take a slightly different approach, please see my comments.

hashpalk:
hey @ericglau, I have tried to follow the approach you suggested. I had to add ES2020.BigInt as it is not available in ES2019 or ES2018.

can you check if I am going in the right direction?

ericglau:
@hashpalk Thanks for following up, this is looking good!  I pushed some refactors, fixed the calculation to handle negative exponents, and added tests.  It should work with es2019 since that contains BigInt.

hashpalk:
hey @ericglau, when I was using BigInt from ES2019, it was saying something like BigInt not found. I was also confused, as BigInt was used in other places. So I had to take it from ES2020.
However, it looks like it is working now after your changes (without ES2020.BigInt). I am guessing it could be some temporary issue with the IDE

packages/core/solidity/src/erc20.ts
```diff
@@ -163,6 +163,32 @@ export function isValidChainId(str: string): boolean {
   return chainIdPattern.test(str);
 }
 
+const UINT256_MAX = BigInt(2) ** BigInt(256) - BigInt(1);
+
+function toUint256(value: string, field: string): bigint {
```
CoveMB, line 168: This function could potentially be moved in a utils file (outside of erc20 specific file)

packages/core/solidity/src/erc20.ts
```diff
@@ -181,6 +207,19 @@ function addPremint(
       const units = integer + decimals + zeroes;
       const exp = decimalPlace <= 0 ? 'decimals()' : `(decimals() - ${decimalPlace})`;
 
+      // Validate base units is within uint256 range
```
CoveMB, line 210: nit: I don't thin the comment is necessary
reading `validatedBaseUnits = toUint256` ~= `Validate base units is within uint256 range`

CoveMB: but it could be added to `toUint256` function doc

packages/core/solidity/src/erc20.ts
```diff
@@ -181,6 +207,19 @@ function addPremint(
...
+      const validatedBaseUnits = toUint256(units, 'premint');
+
+      // Check for potential overflow assuming decimals() = 18
+      const assumedExp = decimalPlace <= 0 ? 18 : 18 - decimalPlace;
```
CoveMB, line 214: maybe we could also capture this logic into a function (up to line 221)

CoveMB (commented):
Looking good, mostly some nits 🌾

CoveMB (approved):
⚡

hashpalk:
@ericglau, @CoveMB thanks for the review. Looks like I do not have the option for merging.

PR #489, opened 2025-03-23
fix: Safari rendering issues with code preview when switching tabs or scrolling

### LINKED ISSUE #466: Safari: Rendering issue on mobile when changing tabs
With Safari browser on mobile, or Safari on desktop with a small window:
1. Start from ERC20 tab:
Image
2. Click Stablecoin tab, then scroll to the right.  Notice the right portion of the code which was previously hidden is rendered incorrectly (it should say ERC20Permit but appears like ERC|mit, and "MyToken" is the name for the ERC20 contract but not the name for the Stablecoin contract.  It appears the display is partially cached by the browser)
Image
3. Or from Stablecoin tab, click Votes before scrolling to the right.  Notice rendering issue on the right, where the content is from the previous ERC20 tab.
Image

---

This appears as a rendering issue only.  If you zoom in or resize the window, or interact with the content, it gets corrected.
The actual code when copied or downloaded is also unaffected.

metanodreamer:
This PR fixes #466  in Safari browsers where the code preview would display incorrect or partially cached content when:
1. Switching between tabs (e.g., from ERC20 to Stablecoin)
2. Scrolling horizontally to view the full code
3. Changing contract settings

The issue was caused by Safari's rendering engine not properly repainting/reflowing the code preview container when content changes occurred in combination with horizontal scrolling. This resulted in showing stale or mixed content from previous states.

Fix:
- Added a force re-render mechanism that temporarily removes and re-adds the code preview element
- Applied the fix to both tab changes and contract option updates
- Set minimal timeout (0ms) to ensure the DOM has a chance to properly clear before re-rendering

The actual contract code and functionality remain unchanged - this only affects how Safari renders the content.

Testing:
1. Open the wizard in Safari (mobile or desktop)
2. Switch between tabs (especially ERC20 and Stablecoin)
3. Scroll horizontally to view full code
4. Verify that code preview always shows correct and complete content
5. Verify that contract features (copy, download, etc.) work as expected

### DIFF
--- a/packages/ui/src/solidity/App.svelte
+++ b/packages/ui/src/solidity/App.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-    import { createEventDispatcher } from 'svelte';
+    import { createEventDispatcher, tick } from 'svelte';
 
     import hljs from './highlightjs';
 
@@ -37,15 +37,18 @@
 
     const dispatch = createEventDispatcher();
 
+    async function allowRendering() {
+      showCode = false;
+      await tick();
+      showCode = true;
+    }
+
     export let initialTab: string | undefined = 'ERC20';
 
     export let tab: Kind = sanitizeKind(initialTab);
     $: {
       tab = sanitizeKind(tab);
-      showCode = false;
-      setTimeout(() => {
-        showCode = true;
-      }, 0);
+      allowRendering();
       dispatch('tab-change', tab);
     };
 
@@ -93,10 +96,7 @@
             throw e;
           }
         }
-        showCode = false;
-        setTimeout(() => {
-          showCode = true;
-        }, 0);
+        allowRendering();
       }
     }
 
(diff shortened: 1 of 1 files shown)


metanodreamer:

metanodreamer:
@ericglau, I have sent the PR. Please review the code

CoveMB:
Looking good thanks for tacking the issue, what do you think about wrapping the duplicated logic into a function maybe called `allowRendering` it would give more context about the usage of setTimeout 

Also have you tried using built in `tick` (to avoid microtasks)?

`
tick().then(() => {
      showCode = true;
 })
`

metanodreamer:
thank you @CoveMB for the suggestion. Using tick() is indeed more idiomatic and better suited for this use case. I have done the following changes:
- Introduced `allowRendering()` function
- Replaced `setTimeout` with `tick()` for better state synchronization
- Applied fix to both tab changes and contract option updates

let me know if anything else needs to be changed.

CoveMB (approved):
✨

metanodreamer:
hi @CoveMB, thank you for the review. I cannot merge the PR. Can you merge it for me?

PR #491, opened 2025-03-24
Support Cairo 2.11.3

immrsd:


packages/core/cairo_alpha/src/utils/version.ts
```diff
@@ -8,8 +8,8 @@ export const contractsVersionTag = `v${contractsVersion}`;
  * Cairo compiler versions.
```
ericglau, line 8: `contractsVersion` above should also use `2.0.0-alpha.0` so that import links work.

packages/core/cairo_alpha/src/utils/version.ts
```diff
@@ -8,8 +8,8 @@ export const contractsVersionTag = `v${contractsVersion}`;
...
  */
 export const edition = '2024_07';
-export const cairoVersion = '2.9.4';
-export const scarbVersion = '2.9.4';
+export const cairoVersion = '2.11.2';
+export const scarbVersion = '2.11.3';
 
 /**
  * Semantic version string representing of the minimum compatible version of Contracts to display in output.
```
ericglau, line 15: Should `compatibleContractsSemver` below be updated to `^2.0.0-alpha.0`?
Although `^2.0.0-alpha` still works because `^2.0.0-alpha` has a lower semver than `^2.0.0-alpha.0`

immrsd: I think we can go with `^2.0.0-alpha.0`. Btw, @ericnordelo are we going to release a final alpha version (`2.0.0-alpha`) or it's unlikely to happen?

If we really want to explicitly exclude the potential final alpha version, we can update the compatible semver to `">=2.0.0-alpha.0 <2.0.0-alpha"`

ericglau: `2.0.0-alpha` is not suitable for a final alpha, because according to semver, `2.0.0-alpha < 2.0.0-alpha.0`

packages/core/cairo_alpha/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.1.0 (2025-03-25)
```
ericglau, line 3: This changelog applies to the alpha NPM package which we aren't planning to release.  Eventually this will be merged into the stable `cairo` folder.  So I think we should mark this as `Unreleased` and accumulate the set of changes for eventual release.
```suggestion
## Unreleased
```

packages/core/cairo_alpha/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@openzeppelin/wizard-cairo-alpha",
   "private": true,
-  "version": "1.0.0",
+  "version": "1.1.0",
```
ericglau, line 4: Not necessary to bump this, since the package is private.  (Although it's fine to do so)

ericglau (approved):

packages/core/cairo_alpha/package.json
ericglau: If we do bump this, it should be according to the planned version when we move this to stable.  Since the Cairo lib will be 2.x and assuming it's a breaking change, this could be `2.0.0`.

Or `2.0.0-alpha.0` for now to match.

packages/core/cairo_alpha/src/utils/version.ts
ericnordelo: > Btw, @ericnordelo are we going to release a final alpha version

I don't think is worth it. We will just release a stable version from the latest alpha after the audits, so every alpha should have a number

packages/core/cairo_alpha/package.json
immrsd: Updated to `2.0.0-alpha.0`

PR #500, opened 2025-03-27
Add ERC1363 to wizard

vittominacori:
Add ERC1363 as available extension for ERC20 tokens.


vittominacori:

packages/ui/src/solidity/ERC20Controls.svelte
```diff
@@ -111,6 +111,14 @@
       </HelpTooltip>
     </label>
 
+    <label class:checked={opts.erc1363}>
+      <input type="checkbox" bind:checked={opts.erc1363}>
+      ERC1363
```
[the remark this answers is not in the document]
vittominacori: What would you suggest?

The [ERC-1363](https://eips.ethereum.org/EIPS/eip-1363) is named "Payable Token" on the official EIP repo.
Would you suggest to change it to "Payable" to keep consistency to the ERC standard or could the label be like "Contract Callback" or something similar?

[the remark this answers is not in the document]
CoveMB: I think "Payable" aligns well with the intent, the official EIP and the other options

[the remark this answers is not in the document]
vittominacori: Yes, but I think even though its name is `Payable Token` it might sound like the Ethereum `payable` keyword (like accept Ether) and could be misunderstood.

I would suggest using `Callback` which could directly connect to ERC1363's `transferAndCall` or `approveAndCall` functionality. Otherwise, if you want to keep payable, we could use `Payable (ERC1363)` for explicitness.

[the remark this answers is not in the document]
CoveMB: What about `Callback Payable` ?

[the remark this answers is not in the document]
CoveMB: ✅

packages/ui/src/solidity/RealWorldAssetControls.svelte
```diff
@@ -119,6 +119,14 @@
       </HelpTooltip>
     </label>
 
+    <label class:checked={opts.erc1363}>
+      <input type="checkbox" bind:checked={opts.erc1363}>
+      Payable Callback
```
Amxx, line 124: I'm not sure id "payable " is the right formulation here. @ernestognw

vittominacori: As discussed above, I initially proposed only `ERC1363`, then modified it based on comments.

I also agree that `payable` could be misunderstood, so I suggested `Callback` to bind to `*AndCall` methods.

Maybe `Callback (ERC1363)` could be better.

Amxx: I would simply say "Callback".

Amxx: Payable would be confused with the idea of sending ether. Also it doesn't apply to approveAndCall well

vittominacori: I renamed to `Callback` because I also believe that `payable` could be misunderstood.

In reality, however, the "payable token" logic also works with "approve", because the `sender` could call `approveAndCall` on a `spender` contract that could then execute a `transferFrom` to complete the payment. As logically happens in the case of a pre-authorization for a credit card or a paypal payment. 

```solidity
function onApprovalReceived(address owner, uint256 value, bytes calldata data) external returns (bytes4) {
    address token = msg.sender;
    IERC20(token).transferFrom(owner, address(this), value);

    // does something

    return this.onApprovalReceived.selector;
}
```

Indeed, for example, if I wanted to use ERC1363 to activate an annual subscription of 100 Tokens/month (so 1200 Tokens/year), I could call `approveAndCall` with an allowance of 1200 Tokens, and the `spender` contract could transfer at the moment only the 100 Tokens needed for the first month, leaving an additional allowance of 1100 Tokens to be used later.

In any case, only `Callback` could be safer to use in this context.

packages/core/solidity/src/erc20.ts
```diff
@@ -26,6 +26,7 @@ export interface ERC20Options extends CommonOptions {
   premint?: string;
   premintChainId?: string;
   mintable?: boolean;
+  erc1363?: boolean;
```
CoveMB, line 29: let's also call this `callback`

CoveMB (commented):
Thanks, let's also add an entry in CHANGELOG.md

packages/core/solidity/src/erc20.ts
Amxx: This is not user facing. I'd keep the erc number to clarify exactly which callback mechanism is used (in case there ever is another one)

packages/core/solidity/src/erc20.test.ts.md
```diff
@@ -971,7 +989,7 @@ Generated by [AVA](https://avajs.dev).
         function supportsInterface(bytes4 interfaceId)␊
             public␊
             view␊
-            override(ERC20Bridgeable, AccessControl)␊
+            override(ERC20Bridgeable, AccessControl, ERC1363)␊
```
Amxx, line 992: I'm curious how ordering is done. Not a blocker (I'd merge like that) but if we can clarify this in a further PR that would be great.

Amxx (approved):

CoveMB (approved):

PR #501, opened 2025-03-28
Fix spacing in UI code display

ericglau:
- Apply https://github.com/OpenZeppelin/contracts-wizard/pull/495 to `cairo_alpha` UI
- Fixes extra whitespace in UI before the contract code, due to whitespace in the `<code>` tag which caused it to look like:
Screenshot 2025-03-28 at 9 27 55 AM


CoveMB (approved):
Thanks 🙏

PR #503, opened 2025-03-28
Disable action buttons when errors present

ericglau:
Disables "Copy to Clipboard" and "Download" buttons in Cairo, Stylus, Stellar.  (This already exists in Solidity)

Testcase: enter some letters in Premint field of ERC20.  Action buttons are disabled.
Screenshot 2025-03-28 at 12 21 00 PM


CoveMB (approved):
LGTM if relevant maybe we could pass the errors to ErrorDisabledActionButtons to display which error(s) is present

ericglau:
Since errors are thrown from core, only one error is present at a time, and it should be highlighted on the left.  The tooltip might disappear if the user clicks somewhere though.

We could consider a way to redirect the user's attention to the relevant input box or redisplay the error tooltip on the left (or prevent it from disappearing)

PR #504, opened 2025-03-28
Add API development server

CoveMB:
This will allow developing around the AI endpoint.

- Add development api server and controller connected to development UI 
- Use Deno with watch mode to run development server.
- Add redis instance mock in development


package.json
```diff
@@ -5,6 +5,8 @@
     "prepare": "wsrun -m prepare",
     "lint": "eslint",
     "dev:ui": "yarn --cwd ./packages/ui dev",
+    "dev:api": "yarn --cwd ./packages/ui dev:api",
+    "dev": "yarn dev:ui & yarn dev:api",
```
ericglau, line 9: This leaves one instance in the background which can't be killed easily.  We could use https://www.npmjs.com/package/concurrently so that it can run and stop both commands simultaneously.

packages/ui/package.json
```diff
@@ -16,8 +17,8 @@
     "@rollup/plugin-replace": "^6.0.0",
     "@rollup/plugin-typescript": "^12.0.0",
     "@types/file-saver": "^2.0.1",
-    "@types/resize-observer-browser": "^0.1.5",
     "@types/node": "^18.0.0",
+    "@types/resize-observer-browser": "^0.1.5",
     "@types/semver": "^7.5.7",
     "autoprefixer": "^10.4.2",
```
ericglau, line 23: We should add deno here so that it gets installed automatically for development:
```suggestion
    "autoprefixer": "^10.4.2",
    "deno": "^2.2.6",
```

README.md
```diff
@@ -13,12 +13,15 @@ Contracts Wizard is a web application to interactively build a contract out of c
...
 
 `packages/core` contains the code generation logic for each language under separately named subfolders. From each language's subfolder:
+
 - Run `yarn test` to run the tests.
 - Run `yarn test:update-snapshots` to update AVA snapshots and run the tests.
 - Run `yarn lint` to run the linter across the codebase (optionally `yarn lint --fix` will automatically fix fixable issues, like formatting issues).
 
-`packages/ui` is the interface built in Svelte. From the `packages/ui` directory, run `yarn dev` to spin up a local server to develop the UI.
-- By default, the local server runs on port 8080. To use another port, set the environment variable `PORT`, for example: `PORT=800 yarn dev`
+`packages/ui` is the interface built in Svelte. From the `packages/ui` directory, run `yarn dev` to spin up a local UI server to develop the UI along side an API server (for edge functions).
```
ericglau, line 21: This is referring to `packages/ui` but the command that spins up both the UI and API are in the root.  We should have the instructions only mention one or the other, to avoid confusion. (Also applies to the bullet points below)

packages/ui/api/ai.ts
```diff
@@ -49,7 +36,7 @@ export default async (req: Request) => {
     ];
 
     const response = await openai.chat.completions.create({
-      model: 'gpt-4-1106-preview',
+      model: getEnvironmentVariableOr('OPEN_AI_MODEL', 'gpt-4o-mini'),
```
ericglau, line 39: For consistency with `OPENAI_API_KEY` name format:
```suggestion
      model: getEnvironmentVariableOr('OPENAI_MODEL', 'gpt-4o-mini'),
```

packages/ui/package.json
CoveMB: I have been having trouble after adding this package like `node/20.11.1/bin/node: --watch: not found
error Command failed with exit code 9.` because node_modules/.bin/deno calls Node instead.
Instead I added install Deno instruction in Readme

packages/ui/package.json
```diff
@@ -37,6 +37,7 @@
     "typescript": "^5.0.0"
   },
   "dependencies": {
+    "concurrently": "^9.1.2",
```
ericglau, line 40: This should be in `devDependencies` instead.

.vscode/exemple.settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
ericglau, line 1: Let's rename this file to `example.settings.json`

package.json
```diff
@@ -5,6 +5,8 @@
     "prepare": "wsrun -m prepare",
     "lint": "eslint",
     "dev:ui": "yarn --cwd ./packages/ui dev",
+    "dev:api": "yarn --cwd ./packages/ui dev:api",
+    "dev": "concurrently --kill-others-on-fail --names \"client,server\" --prefix-colors \"magenta.bold,green.bold\" \"yarn dev:ui\" \"yarn dev:api\"",
```
ericglau, line 9: Perhaps name as ui,api to make it clearer
```suggestion
    "dev": "concurrently --kill-others-on-fail --names \"ui,api\" --prefix-colors \"magenta.bold,green.bold\" \"yarn dev:ui\" \"yarn dev:api\"",
```

packages/ui/package.json
```diff
@@ -16,10 +17,11 @@
     "@rollup/plugin-replace": "^6.0.0",
     "@rollup/plugin-typescript": "^12.0.0",
     "@types/file-saver": "^2.0.1",
-    "@types/resize-observer-browser": "^0.1.5",
     "@types/node": "^18.0.0",
+    "@types/resize-observer-browser": "^0.1.5",
     "@types/semver": "^7.5.7",
     "autoprefixer": "^10.4.2",
+    "concurrently": "^9.1.2",
```
ericglau, line 24: I just noticed since `concurrently` is run from the root dir, it would be more correct to move this to the root package.json's devDependencies instead.

ericglau (approved):
Just a few more comments, the rest LGTM. Thanks!

ericglau (approved):

PR #505, opened 2025-03-28
Plat 6355 provide current generated contract as context to open ai

### LINKED ISSUE #442: AI Assistant: Provide current contract code as context
The AI Assistant has the currently selected options as context, but is not the actual contract code which results from the current options.
Provide the current contract code as context, so that it can answer direct questions related to the code that the user sees on screen.

CoveMB:
Note: to review after merging API server changes to limit scope of changes

Fixes #442 

Add current contract code to payload sent to /ai API to be included in the chat with AI.

Test cases ask AI question that requires code to answer 
Screenshot 2025-03-29 at 2 30 26 PM


packages/ui/api/ai.ts
```diff
@@ -11,33 +11,46 @@ import {
 import { getRedisInstance } from './services/redis.ts';
 import { getOpenAiInstance } from './services/open-ai.ts';
 import { getEnvironmentVariableOr } from './utils/env.ts';
+import type { Chat } from '../src/types.ts';
```
CoveMB, line 14: the type file will be moved in next MR

yarn.lock
```diff
@@ -888,6 +888,11 @@
   resolved "https://registry.yarnpkg.com/@tsconfig/node16/-/node16-1.0.4.tgz#0b92dcc0cc1c81f6f306a381f28e31b1a56536e9"
   integrity sha512-vxhUy4J8lyeyinH7Azl1pdd43GJhZH/tP2weN8TntQblOY+A0XbT8DJk1/oCPuOOyg/Ja757rG0CgHcWC8OfMA==
 
+"@tsconfig/svelte@^5.0.4":
```
ericglau, line 891: Is this needed?  (And this isn't the svelte version we are using)

PR #508, opened 2025-04-02
Plat 6356 refactor ai assistant code to prepare for adding other

CoveMB:
Refactor AI assistant code to expand to other languages
- Import language contract options in type file (as Deno was having trouble capturing those type either when KinderedOptions or capturing the type of interfaces with extending from an other interface)
- Create typing for AI assistant API function definition and link it to the language definition 
- enforce definition with typing (use `anyOf` for mixed types)(use temporarily Omit for missing fields)
- use contract kind instead of mapping name
- adjust Wiz component to not use type conversion as it is not needed any more
- pass a language parameter to API to use different language function definitions as context
- move UI component in common folder


packages/ui/api/ai-assistant/function-definitions/solidity.ts
```diff
@@ -1,79 +1,27 @@
...
     type: 'object',
     properties: {
-      name: repeatedOptions.name,
-      symbol: repeatedOptions.symbol,
-      burnable: repeatedOptions.burnable,
-      pausable: repeatedOptions.pausable,
+      ...addFunctionProperties(['name', 'symbol', 'burnable', 'pausable', 'mintable', 'access', 'upgradeable', 'info']),
       premint: {
-        type: 'number',
+        type: 'string',
```
CoveMB, line 12: need to check how the AI assistant handle string representation of numbers

packages/ui/api/ai-assistant/types/function-definition.ts
```diff
@@ -0,0 +1,53 @@
+import type { AllLanguageContractOptions } from './languages.ts';
+
+type AiFunctionCallPrimaryType<TType> = TType extends string
+  ? 'string'
+  : TType extends number
+    ? 'number'
```
CoveMB, line 6: ..otherwise will set this as string

packages/ui/src/common/Wiz.svelte
```diff
@@ -36,15 +42,7 @@
...
-  const nameMap = {
-    erc20: 'ERC20',
-    erc721: 'ERC721',
-    erc1155: 'ERC1155',
-    stablecoin: 'Stablecoin (Experimental)',
-    realworldasset: 'Real-World Asset (Experimental)',
-    governor: 'Governor',
-    custom: 'Custom',
-  };
+  const experimentalContracts = ['Stablecoin', 'RealWorldAsset'];
```
ericglau, line 45: Can we make `experimentalContracts` and `sampleMessages` (below) mapped based on the currently chosen `language`?

Reason is that we may want these to be different depending on language.

packages/ui/api/ai.ts
```diff
@@ -1,23 +1,18 @@
...
-  chatId: string;
+const getFunctionsContext = (language: SupportedLanguage) => {
+  const functionPerLanguages: Record<SupportedLanguage, Record<string, AiFunctionDefinition>> = {
+    solidity: solidityFunctions,
+  };
+
+  return Object.values(functionPerLanguages[language] ?? {});
 };
 
 const buildAiChatMessages = (request: AiChatBodyRequest): Chat[] => {
```
ericglau, line 18: It may help to provide the language as context to the system prompt below, to tell the AI assistant that the current contract code is for a specific smart contract language.

packages/ui/api/ai-assistant/types/languages.ts
```diff
@@ -0,0 +1,38 @@
...
+  ERC20: { kind: 'ERC20' } & SolidityCommonOptions & SolidityERC20Options;
+  ERC721: { kind: 'ERC721' } & SolidityCommonOptions & SolidityERC721Options;
+  ERC1155: { kind: 'ERC1155' } & SolidityCommonOptions & SolidityERC1155Options;
+  Stablecoin: { kind: 'Stablecoin' } & SolidityCommonOptions & SolidityERC20Options & SolidityStablecoinOptions;
+  RealWorldAsset: { kind: 'RealWorldAsset' } & SolidityCommonOptions & SolidityERC20Options & SolidityStablecoinOptions;
+  Governor: { kind: 'Governor' } & SolidityCommonOptions & SolidityGovernorOptions;
+  Custom: { kind: 'Custom' } & SolidityCommonOptions & SolidityCustomOptions;
+}
+
+export type { SolidityCommonOptions };
```
ericglau, line 30: Would it be simpler to just directly use `export interface SolidityCommonOptions {` on line 14?

packages/ui/api/ai-assistant/function-definitions/shared.ts
```diff
@@ -0,0 +1,75 @@
+import type { AiFunctionPropertyDefinition } from '../types/function-definition.ts';
+import type { SolidityCommonOptions } from '../types/languages.ts';
+
+export const commonFunctionDescription = {
```
ericglau, line 4: Can `commonFunctionDescription` be moved into `solidity.ts` instead?  The possible options vary depending on the language, for example Cairo does not support `access: 'managed'`

packages/ui/api/ai-assistant/function-definitions/shared.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  burnable: {
+    type: 'boolean',
+    description: 'Whether token holders will be able to destroy their tokens',
+  },
+
+  pausable: {
+    type: 'boolean',
+    description:
+      'Whether privileged accounts will be able to pause the functionality marked as whenNotPaused. Useful for emergency response.',
```
ericglau, line 53: We can generalize the wording here so that it can apply for all languages. 
```suggestion
      'Whether privileged accounts will be able to pause specific functionality. Useful for emergency response.',
```

packages/ui/src/cairo/App.svelte
```diff
@@ -1,151 +1,136 @@
 <script lang="ts">
```
ericglau, line 1: Just a reminder to revert this file if there are no changes.

ericglau (approved):
Looks great!

PR #514, opened 2025-04-09
Plat 6357 add cairo ai assistant

CoveMB:
- Add AI assistant for Cairo and Cairo Alpha


ericglau (approved):
LGTM, thanks!
@ericnordelo @immrsd It would be great if you could review this as well, and note these should be kept up to date as new options/features are added to the Cairo (Stable/Alpha) Wizards.

ericnordelo (approved):
Great addition! Looks good to me!

PR #515, opened 2025-04-09
Plat 6359 add stylus ai assistant

CoveMB:
- Add backend AI assistant support for Stylus
- Add Wiz UI component in Stylus page


ericglau (approved):
LGTM, thanks!
@0xNeshi @bidzyyys @qalisander It would be great if you could review this as well, and note these should be kept up to date as new contracts/options are added to the Stylus Wizard.

packages/ui/src/stylus/App.svelte
```diff
@@ -113,9 +117,21 @@
...
+  };
 </script>
 
 <div class="container flex flex-col gap-4 p-4 rounded-3xl">
+  <WizStylus
+    language="stylus"
+    bind:currentOpts={opts}
+    bind:currentCode={code}
+    on:function-call-response={applyFunctionCall}
+    sampleMessages={['Make a token with supply of 10 million', 'What does mintable do?']}
```
ericglau, line 133: These sample messages aren't relevant for Stylus because Stylus Wizard currently does not support premint or mintable.

@0xNeshi @bidzyyys @qalisander Any suggestions on other sample messages that we could provide?
This would show up at the start of the AI chat session, for example:
Screenshot 2025-04-09 at 4 30 11 PM

packages/ui/api/ai-assistant/function-definitions/stylus-shared.ts
```diff
@@ -0,0 +1,22 @@
+import type { AiFunctionPropertyDefinition } from '../types/function-definition.ts';
+import type { StylusCommonContractOptions } from '../types/languages.ts';
+
+export const stylusCommonFunctionDescription = {
+  access: {
+    type: 'boolean',
+    enum: [false],
```
0xNeshi, line 7: What does this line mean for `access`?

packages/ui/api/ai-assistant/function-definitions/stylus.ts
```diff
@@ -0,0 +1,61 @@
...
+import { addFunctionPropertiesFrom } from './shared.ts';
+import { stylusCommonFunctionDescription } from './stylus-shared.ts';
+
+export const stylusERC20AIFunctionDefinition = {
+  name: 'ERC20',
+  description: 'Make a fungible token per the ERC-20 standard',
+  parameters: {
+    type: 'object',
+    properties: {
+      ...addFunctionPropertiesFrom(stylusCommonFunctionDescription, ['name', 'burnable', 'access', 'info']),
```
0xNeshi, line 11: `access` contracts are not yet enabled for Stylus, should this item be removed from the array?

packages/ui/src/stylus/App.svelte
0xNeshi: We could maybe comment the current message out and keep it in code for future reference.

Do any of you have some statistic on which contract features are most often generated through wizard? It would make sense to create messages related to those features.

If any placeholder will do, we could temporarily use something like:
```ts
sampleMessages={['Make a token with flash loan capability', 'What is a flash loan?']}
```

packages/ui/src/stylus/ERC20Controls.svelte
```diff
@@ -8,7 +8,7 @@
   import InfoSection from './InfoSection.svelte';
   import { error } from '../common/error-tooltip';
 
-  export const opts: Required<KindedOptions['ERC20']> = {
+  export let opts: Required<KindedOptions['ERC20']> = {
```
0xNeshi, line 11: nit: why the change?

CoveMB: to allow the parent component updating the variable (but we will update nested values instead and change all exported opts as const)

packages/ui/api/ai-assistant/function-definitions/stylus-shared.ts
CoveMB: That the AI assistant should only/always return false (as it is not yet supported)

packages/ui/api/ai-assistant/function-definitions/stylus.ts
CoveMB: we are keeping to have all proprieties represented but allowing AI assistant to only return false

0xNeshi (approved):
Only [this comment](https://github.com/OpenZeppelin/contracts-wizard/pull/515/files#r2036100771) left to address, but other than that looks good :+1:

PR #523, opened 2025-04-23
Address feedback for Account

ernestognw:
#### Description

Just fixing a couple of nits:

- The default signer is `false`, which affects the discoverability of the "Signers" menu
- In ERC7739, there's a small imprecision about accounts owned by the same external key, not necessarily an EOA
- Changing Hooks to Hooked to make it clear that the account only supports one hook


ericglau (approved):
LGTM, thanks!
To fix the tests, please update snapshots by running `yarn test:update-snapshots` from `packages/core/solidity`

PR #525, opened 2025-04-24
feat: positioned the assistant button in such a way that it wont be affected in different screen sizes

### LINKED ISSUE #524: Adjust default window height
On MacOS, the default height is too tall, which causes the "AI Assistant" button to be cut off at the bottom.

Example:
Image

We should adjust the height so that all of the content fits by default.

muhammedsirajudeen:
Fixes #524 

#### Description

The AI button was getting cutoff in different screen sizes so I adjusted the value so that it would be there regardless of the screen sizes thus improving ux

#### Image
Screenshot from 2025-04-24 15-29-24

### DIFF
--- a/packages/ui/src/common/Wiz.svelte
+++ b/packages/ui/src/common/Wiz.svelte
@@ -326,7 +326,7 @@
     </button>
   {:else}
     <button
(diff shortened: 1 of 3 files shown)


muhammedsirajudeen:

muhammedsirajudeen:
Hey i have brought everything to the viewport as you mentioned.
image

Do let me know if any other changes are required.

ericglau (approved):
Thanks, this looks great! I've adjusted the calculation to use a specific pixel height diff so that it remains stable even when resizing the window.

PR #527, opened 2025-04-25
Add multisig to Account Wizard

ernestognw:
### Description

Adding MultisigERC7913Signer and MultisigERC7913 recently added to community contracts here: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/116


packages/core/solidity/src/account.ts
```diff
@@ -175,6 +176,23 @@ function addERC7579Modules(c: ContractBuilder, opts: AccountOptions): void {
...
+    case 'MultisigWeighted':
+      c.addFunctionCode(
+        `_setSignerWeights(${functions.setSignerWeights.args.map(({ name }) => name).join(', ')});`,
+        functions.setSignerWeights,
+      );
+    // eslint-disable-next-line no-fallthrough
+    case 'Multisig':
+      c.addFunctionCode(`_addSigners(${functions.addSigners.args[0]?.name});`, functions.addSigners);
+      c.addFunctionCode(`_removeSigners(${functions.removeSigners.args[0]?.name});`, functions.removeSigners);
+      c.addFunctionCode(`_setThreshold(${functions.setThreshold.args[0]?.name});`, functions.setThreshold);
```
ericglau, line 190: Can these use non-null assertions since we know that args[0] is available?
```suggestion
      c.addFunctionCode(`_addSigners(${functions.addSigners.args[0]!.name});`, functions.addSigners);
      c.addFunctionCode(`_removeSigners(${functions.removeSigners.args[0]!.name});`, functions.removeSigners);
      c.addFunctionCode(`_setThreshold(${functions.setThreshold.args[0]!.name});`, functions.setThreshold);
```

packages/core/solidity/src/signer.ts
```diff
@@ -19,13 +19,28 @@ export function addSigner(c: ContractBuilder, signer: SignerOptions): void {
...
+
+  switch (signer) {
+    case 'Multisig':
+      c.addFunctionCode(`_addSigners(${fn.args?.[0]?.name});`, fn);
+      c.addFunctionCode(`_setThreshold(${fn.args?.[1]?.name});`, fn);
+      break;
+    case 'MultisigWeighted':
+      c.addFunctionCode(`_addSigners(${fn.args?.[0]?.name});`, fn);
+      c.addFunctionCode(`_setSignerWeights(${fn.args?.[0]?.name}, ${fn.args?.[1]?.name});`, fn);
+      c.addFunctionCode(`_setThreshold(${fn.args?.[2]?.name});`, fn);
```
ericglau, line 31: Similar to my other comment, can these use non-null assertions?

It may also be useful to use some local variables to make this more readable e.g.
```
const signers = fn.args!.[0]!.name;
const weights = fn.args!.[1]!.name;
const threshold = fn.args!.[2]!.name;
...
```

packages/ui/src/solidity/AccountControls.svelte
```diff
@@ -176,6 +176,29 @@
...
+      Multisig Weighted
+      <HelpTooltip
+        link="https://docs.openzeppelin.com/community-contracts/0.0.1/multisig-account#multisignererc7913weighted"
+      >
+        Extension of ERC-7913 multisignature validation that supports weighted signatures. Each signer can be assigned a
+        different weight, enabling flexible governance schemes where some signers have more influence than others. The
+        contract maintains a set of authorized signers and validates that the total weight of valid signatures meets the
+        threshold requirement. Signers are encoded as `verifier || key` and must be ordered by their keccak256 hash to
+        prevent duplicates.
+      </HelpTooltip>
```
ericglau, line 200: These descriptions are getting quite long.  We should aim to keep these succinct, with just enough information to help the user differentiate the different options.  Users can click the links for more details (and ensure the linked docs have all of the relevant info).

yarn.lock
```diff
@@ -7,11 +7,6 @@
   resolved "https://registry.yarnpkg.com/@alloc/quick-lru/-/quick-lru-5.2.0.tgz#7bf68b20c0a350f936915fcae06f58e32007ce30"
   integrity sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==
 
-"@axelar-network/axelar-gmp-sdk-solidity@^6.0.6":
```
ericglau, line 10: Nice.  (Builds were previously failing intermittently due to this dependency)

packages/ui/src/solidity/AccountControls.svelte
ernestognw: Fair, I simplified these descriptions 😄

yarn.lock
ernestognw: Interesting. I never noticed we should've set it in `peerDependencies`. I updated that on [community contracts](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/123)

packages/ui/api/ai-assistant/function-definitions/solidity.ts
```diff
@@ -209,7 +209,7 @@ export const solidityAccountAIFunctionDefinition = {
       signer: {
         anyOf: [
           { type: 'boolean', enum: [false] },
-          { type: 'string', enum: ['ECDSA', 'ERC7702', 'P256', 'RSA'] },
+          { type: 'string', enum: ['ECDSA', 'ERC7702', 'P256', 'RSA', 'Multisig', 'MultisigWeighted'] },
         ],
         description: 'Defines the signature verification algorithm used by the account to verify user operations.',
```
CoveMB, line 214: Maybe we could add a small description of each signature in description field? (similar to what is found in tooltip)

ericglau (approved):

packages/ui/api/ai-assistant/function-definitions/solidity.ts
ernestognw: Is there a way to provide descriptions for an enum? I would agree on providing more context about what they mean

ericglau: It should be possible to just directly use the common `description` field for `signer` to describe all of the enums as well.  The LLM should be able to understand it as context.

ernestognw: https://github.com/OpenZeppelin/contracts-wizard/pull/527/commits/623bd092b2fe7632c0d908464d00aec4c93f03d4 must do the job

ericglau (approved):

PR #528, opened 2025-04-25
Use changesets for releases

ericglau:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Check changesets status
+        if: steps.changesets.outputs.hasChangesets == 'true'
+        run: |
+          echo "Changesets found. Merge Prepare Release PR before publishing."
+          exit 1
+      - name: Check publish status
+        if: steps.changesets.outputs.published == 'false'
+        run: |
+          echo "Publish failed. Check the logs for more details."
+          exit 1
```
tirumerla, line 43: Instead of failing changesets status we probably can skip it if it's a prepare release pr 

For publish status `steps.changesets.outputs.published == 'false' && steps.changesets.outputs.hasChangesets == 'false'` ? Isn't for release pr published will be false?

scripts/release/publish.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+yarn install --frozen-lockfile
+changeset publish
+git push --follow-tags
```
tirumerla, line 7: nit: EOL

tirumerla (approved):
lgtm, minor comments

.github/workflows/publish.yml
ericglau: Running this workflow means that we specifically want to publish, so it seems useful for any other condition to throw an error.  Particularly if it created a Prepare Release PR, it could indicate that there were some unexpected changesets when we wanted to publish.

> For publish status steps.changesets.outputs.published == 'false' && steps.changesets.outputs.hasChangesets == 'false' ? Isn't for release pr published will be false?

Yes these are all correct.  The purpose of these is to give useful errors to tell us what went wrong.  Here is a table summarizing the possible states and what workflow status we want to show.  I think the existing conditions already handle these.

| hasChangesets | published | Workflow status       | Notes                                   |
|---------------|-----------|-----------------|-----------------------------------------|
| false         | false     |  ❌ Error: Publish failed   | No changesets, but publish failed  |
| false         | true      | ✅  Success      | Published, so must not have changesets |
| true          | false     |  ❌ Error: Changesets found       | Has changesets, so must not be published |
| true          | true      | Does not handle this   | Invariant: This should not be possible   |

PR #531, opened 2025-04-28
Stellar v0.2.0

ozgunozerk:
Adds NonFungible module support


packages/ui/api/ai-assistant/function-definitions/stellar.ts
```diff
@@ -26,3 +26,41 @@ export const stellarFungibleAIFunctionDefinition = {
     additionalProperties: false,
   },
 } as const satisfies AiFunctionDefinition<'stellar', 'Fungible'>;
+
+// TODO: fix the errors below
```
CoveMB, line 30: after running yarn prepare I don't see any error in this file, if you get any I can help resolving them

ozgunozerk: this was a leftover, thanks for pointing that out! All works on my end as well 👍

packages/core/stellar/src/non-fungible.test.ts.md
```diff
@@ -0,0 +1,1268 @@
...
+    impl NonFungibleToken for MyToken {␊
+        type ContractType = Base;␊
+    ␊
+        fn owner_of(e: &Env, token_id: u32) -> Address {␊
+            Self::ContractType::owner_of(e, token_id)␊
+        }␊
+    ␊
+        fn transfer(e: &Env, from: Address, to: Address, token_id: u32) {␊
+            Self::ContractType::transfer(e, &from, &to, token_id);␊
+        }␊
```
ericglau, line 38: NonFungible's convention of using `Self::ContractType::transfer` is different from Fungible which has `fungible::transfer`.  Is there an important distinction between these conventions?  Should they be made consistent?

ericglau (commented):
We should also update the required library version in `packages/core/stellar/src/utils/version.ts`.

Note: if this will require the full release of Stellar Contracts 0.2.0, then we should hold off with merging this until that is available.

ozgunozerk:
@brozorec can you check `upgradeable` extension addition?

ozgunozerk:
I don't exactly know why Solidity tests are failing, but I triple checked the changed files, and I didn't touch any solidity files (or general files), so I assume the failure of the test is not related to this PR?

packages/ui/src/stellar/NonFungibleControls.svelte
```diff
@@ -0,0 +1,107 @@
...
+  </div>
+
+  <label class="labeled-input">
+    <span class="flex justify-between pr-2">
+      Premint
+      <HelpTooltip>Generate initial NFTs with given IDs for the owner.</HelpTooltip>
+    </span>
+    <input bind:value={opts.premint} use:error={errors?.premint} placeholder="0" />
+  </label>
+</section>
```
ericglau, line 58: I would suggest removing this option for the following reasons:

In the Fungible case, premint allocates a certain number of tokens to the recipient.  In this Non Fungible case, premint allocates only one token of the given ID.  This is slightly confusing and could be misused, since it's not immediately obvious that the value should represent an ID rather than an amount in this case.

The current setup also only allows one value, not a series of IDs.  It doesn't seem a useful pattern to support only preminting one specific ID.

packages/ui/src/stellar/MintableSection.svelte
```diff
@@ -0,0 +1,59 @@
...
+    mintable = false;
+  }
+
+  $: previousConsecutive = consecutive;
+</script>
+
+<ExpandableToggleRadio
+  label="Minting"
+  bind:value={mintable}
+  defaultValue={true}
```
ericglau, line 28: When enabling "Minting", notice that "Non-Sequential" is selected by default.  Is this the expected default?  If so, we could change the UI order so that "Non-Sequential" is a the top.

Alternatively, instead of having two radio buttons for "Sequential" and "Non-Sequential", what do you think of just using a checkbox for "Sequential" as a sub-option instead?

Something like:
Screenshot 2025-04-30 at 3 14 08 PM
(see Solidity's ERC721's Mintable and Auto Increment Ids for example)

packages/ui/src/stellar/MintableSection.svelte
```diff
@@ -0,0 +1,59 @@
...
+
+  let previousConsecutive = consecutive;
+  function handleSequentialChange(value: boolean) {
+    sequential = value;
+    if (!mintable) {
+      mintable = true;
+    }
+  }
+
+  $: if (consecutive !== previousConsecutive && consecutive) {
```
ericglau, line 17: There's a minor issue with the reactivity here.  If "Minting" and "Sequential" are enabled, then you uncheck "Minting", notice that "Sequential" is still selected.

ericglau:
- Solidity tests issue is not related to this PR. Opened a separate issue https://github.com/OpenZeppelin/contracts-wizard/issues/534
- Reminder to test manually by enabling all options and ensure the generated code compiles successfully on a local project.

packages/core/stellar/src/add-upgradeable.ts
```diff
@@ -0,0 +1,33 @@
+import { getSelfArg } from './common-options';
+import type { ContractBuilder } from './contract';
+import { defineFunctions } from './utils/define-functions';
+
+export function addUpgradeable(c: ContractBuilder) {
```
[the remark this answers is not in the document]
ozgunozerk: They are both extensions and additional traits in our case, so if we are rename it to `set-upgradable`, then I think we also have to rename `add-pausable` to `set-pausable`. Then another discussion arises for regular extensions and utility extensions. I'd like to keep it as `add-upgradeable` tbh.

packages/ui/src/stellar/MintableSection.svelte
ozgunozerk: > When enabling "Minting", notice that "Non-Sequential" is selected by default. Is this the expected default? If so, we could change the UI order so that "Non-Sequential" is a the top.

Yes, it is the default behavior. Good suggestion.

> Alternatively, instead of having two radio buttons for "Sequential" and "Non-Sequential", what do you think of just using a checkbox for "Sequential" as a sub-option instead?

My first thought was to be explicit on these, but after giving it some more thought, I'm inclining towards this approach, will do! Thanks 🙏

packages/ui/src/stellar/MintableSection.svelte
ozgunozerk: thanks, great catch!

packages/ui/src/stellar/MintableSection.svelte
```diff
@@ -0,0 +1,63 @@
...
+>
+  <div class="checkbox-group">
+    <label class:checked={!sequential && mintable}>
+      <input
+        type="radio"
+        name="minting-mode"
+        value="non_sequential"
+        checked={!sequential && mintable && !consecutive}
+        on:change={() => handleSequentialChange(false)}
+        disabled={consecutive || !mintable}
```
ericglau, line 44: Minor preference, we don't necessarily need to disable the radio buttons if mintable is false.  This way, clicking them will cause mintable to become true.
```suggestion
        disabled={consecutive}
```

packages/ui/src/stellar/MintableSection.svelte
```diff
@@ -0,0 +1,63 @@
...
+    </label>
+
+    <label class:checked={sequential}>
+      <input
+        type="radio"
+        name="minting-mode"
+        value="sequential"
+        checked={sequential && !consecutive}
+        on:change={() => handleSequentialChange(true)}
+        disabled={consecutive || !mintable}
```
ericglau, line 57:
```suggestion
        disabled={consecutive}
```

packages/core/stellar/src/non-fungible.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+testNonFungible('non-fungible full', {
+  access: 'ownable',
+  burnable: true,
+  mintable: true,
+  pausable: true,
+  enumerable: true,
+  consecutive: true,
+  sequential: true,
+});
```
ericglau, line 77: Should this throw an error if incompatible options are enabled together?  For example, enumerable and consecutive.  Or consecutive and mintable non-sequential.

(Note if we do throw an error, the UI would need to pass through the error to display using `use:error` in the relevant input fields.)

packages/ui/src/stellar/NonFungibleControls.svelte
```diff
@@ -0,0 +1,99 @@
...
+  export let opts: Required<KindedOptions['NonFungible']> = {
+    kind: 'NonFungible',
+    ...nonFungible.defaults,
+    info: { ...infoDefaults },
+  };
+
+  export let errors: undefined | OptionsErrorMessages;
+
+  // Handler functions for checkbox changes
+  function handleConsecutiveChange(value: boolean) {
```
ericglau, line 20: If you use the AI assistant and tell it to enable everything for Non Fungible, notice that all of the incompatible options are chosen together (e.g. consecutive and enumerable and mintable).  This cannot be avoided even if we adjust the prompts/types for the AI assistant, since it could still force things to be enabled if requested by the user.

This issue occurs because when the AI assistant's options are applied, it does not trigger these functions through `on:change`.

Possible ways to address this could be:
1. Avoid relying on `on:change`, and only use pure reactivity to enable/disable mutually exclusive options.  This would apply to this page and also MintableSection.svelte.
2. Detect when incompatible options are selected (see my other comment) and surface these errors in the UI.

ericglau: The above could also occur if someone uses the `@openzeppelin/wizard-stellar` package and calls the TypeScript API directly to enable all of these fields programmatically.

Note: If selecting mutually exclusive options isn't technically wrong/unsafe (but just discouraged), then perhaps it's fine as is.

packages/core/stellar/src/non-fungible.test.ts
ozgunozerk: Yes, absolutely! We have ensured by Rust's type safety that when conflicting options are enabled together, the code won't even compile, but throwing an error here in the wizard would be super helpful as well. 

Tried to address this in: [fc451ca](https://github.com/OpenZeppelin/contracts-wizard/pull/531/commits/fc451cab41c311dda9a0882157ab59a962134c80)

packages/ui/src/stellar/NonFungibleControls.svelte
ozgunozerk: I see, `on:change` only gets triggered when the changes are done in the UI, but there are other ways to interact with the options, hence `on:change` is not sufficient in these cases to ensure incompatibilities. 

My experience with TypeScript is not vast, and it's been years since I last interacted with it. Tried my best to address the issues in: [fc451ca](https://github.com/OpenZeppelin/contracts-wizard/pull/531/commits/fc451cab41c311dda9a0882157ab59a962134c80), and took some help from AI as well :) 

Still, I'm not super confident on the solution here whether it is following the best practices. So, please feel free to take the reins from here 🙏

packages/ui/src/stellar/MintableSection.svelte
ozgunozerk: If that is not a deal-breaker, I'd like to keep them as is :)

ericglau: It would differ a bit from other pages, but I think this is fine.

ericglau (commented):
The new error messages work great even with the AI assistant.  Thanks for the changes!

Just leaving a checklist of a few remaining things.  I can work on these.
- [x] Enabling Consecutive and Pausable together causes an error in the console:
```
contract.ts:202 Uncaught (in promise) Error: Trait 'NonFungibleBurnable' does not exist and cannot be overridden.
    at Kg.overrideAssocType (contract.ts:202:13)
    at non-fungible.ts:229:7
    at Sb (non-fungible.ts:105:5)
    at build-generic.ts:19:14
    at e.$$.update (App.svelte:76:20)
```
- [x] Enabling Minting does not auto-select Non-Sequential
- [x] Still need to update the version in version.ts, and wait until 0.2.0 is released before we merge this

brozorec (changes requested):
Other issues:
- [ ] When “Consecutive” is selected and then “Pausable” (or vice-versa), the UI is not updated
- [x] When switching between "Sequential" and "Non-Sequential" minting, the UI is not updated
- [x] The anchor "#nonfungible" in the url is not working, always lands on "Fungible"

ericglau (approved):

ozgunozerk:
@ericglau thanks a lot for your effort, inspected all the commits and learned some TypeScript tricks, thanks for that as well 🚀

brozorec (approved):

packages/core/stellar/src/non-fungible.ts
```diff
@@ -0,0 +1,395 @@
...
+  // 3. uncomment the section below:
+  // 4. uncomment `get_owner_token_id` and `get_token_id` from `enumerableFunctions` definition
+  /*
+  c.addFunction(nonFungibleEnumerableTrait, enumerableFunctions.total_supply);
+  c.addFunction(nonFungibleEnumerableTrait, enumerableFunctions.get_owner_token_id);
+  c.addFunction(nonFungibleEnumerableTrait, enumerableFunctions.get_token_id);
+  */
+}
+
+function addConsecutive(c: ContractBuilder, burnable: boolean, access: Access) {
```
ericglau, line 213: Looks like burnable is no longer used here
```suggestion
function addConsecutive(c: ContractBuilder, access: Access) {
```

packages/core/stellar/src/non-fungible.ts
```diff
@@ -0,0 +1,395 @@
...
+  if (allOpts.burnable) {
+    addBurnable(c, allOpts.pausable);
+  }
+
+  if (allOpts.enumerable) {
+    addEnumerable(c, allOpts.burnable);
+  }
+
+  if (allOpts.consecutive) {
+    addConsecutive(c, allOpts.burnable, allOpts.access);
```
ericglau, line 105:
```suggestion
    addConsecutive(c, allOpts.access);
```

packages/core/stellar/src/non-fungible.ts
```diff
@@ -226,48 +225,31 @@ function addConsecutive(c: ContractBuilder, burnable: boolean, pausable: boolean
...
-      if (pausable) {
-        c.addFunctionTag(enumerableFunctions.sequential_mint, 'when_not_paused');
-      }
+      requireAccessControl(c, undefined, enumerableFunctions.sequential_mint, access);
     } else {
       c.addFreeFunction(enumerableFunctions.non_sequential_mint);
-
-      if (pausable) {
-        c.addFunctionTag(enumerableFunctions.non_sequential_mint, 'when_not_paused');
-      }
```
ericglau: @brozorec Can you describe why we don't want/need these mint functions (and `batch_mint` in `addConsecutive` above) to be also gated by `when_not_paused`?

Currently with a single owner, perhaps this would not add extra security because the function is restricted to the owner who also has access to pause/unpause.

But when we expand to support role-based access control, there could be different roles for pausing and calling mint functions.

brozorec: > Currently with a single owner, perhaps this would not add extra security because the function is restricted to the owner who also has access to pause/unpause.

@ericglau yes, this was my reasoning to remove them, but I hear your point, @ozgunozerk also raised it. Will add them back 👍

PR #537, opened 2025-05-01
Add Account compilation tests and simplify community contracts imports

ernestognw:
Adds and fixes compilation tests for Accounts, also simplifies the imports of community contracts from `@openzeppelin/community-contracts/contracts` to only `@openzeppelin/community-contracts`. This is the current imports we're recommending on the documentation.


packages/core/solidity/remappings.txt
```diff
@@ -0,0 +1,2 @@
+@openzeppelin/community-contracts/=node_modules/@openzeppelin/community-contracts/contracts/
+@openzeppelin/contracts-upgradeable/=node_modules/@openzeppelin/contracts-upgradeable/
```
ericglau, line 2: Why is remapping necessary for `@openzeppelin/contracts-upgradeable`?

ernestognw: tbh not sure but made erc20 compilation tests fail. I think the remappings task might be overriding some default remappings (?)

cost/benefit i thought it was fine to just add the remapping to avoid the rabbit hole 😆

ericglau (approved):

PR #543, opened 2025-05-09
Ignore version workflow if only UI changes present

ericglau:
If only the UI package was changed, ignore the version workflow which updates the Prepare Release PR since the core packages don't need to be released.


CoveMB (approved):

tirumerla (approved):
lgtm

PR #547, opened 2025-05-13
Escape encoding of special character to avoid crash of bto

### LINKED ISSUE #545: Disable Open in Remix button if contract has Unicode characters
When using unicode characters (for example in a token name), Open in Remix doesn't work and causes a console error:
```
Uncaught (in promise) InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at remix.ts:3:34
    at Blob.type (App.svelte:197:17)
```
We currently encode ASCII to base64, because Remix takes a [base64 encoded](https://remix-ide.readthedocs.io/en/latest/locations.html#load-an-encoded-base64-string-into-a-sol-file-in-the-editor) string and decodes it to ASCII. 

If we convert a Unicode string to UTF-8 before encoding to base64, Remix would not decode it correctly as UTF-8.

Therefore, for now if Unicode characters are present in the contract, we should disable the Open in Remix button and give an appropriate message when hovering over it, something like "Open in Remix is not supported with Unicode characters"

CoveMB:
If a contract contained special characters opening it with remix would fail as `btoa` would not support special characters.
The following change will escape special characters before encoding.

This will ensure that the app does not crash in the presence of special characters.
A change has also been proposed on remix side to support [decoding of the escaped special characters](https://github.com/ethereum/remix-project/pull/6023).

Fixes #545


CoveMB:
[Got merged](https://github.com/ethereum/remix-project/pull/6023) on Remix's side

ericglau (approved):
LGTM! I think we'd still need to wait for the next Remix release before merging.

PR #549, opened 2025-05-14
Feat/59 add natspec descriptions

### LINKED ISSUE #59: Add NatSpec description to generated functions

mrscottyrose:
Adds NatSpec descriptions to generated functions in the Contracts Wizard. This improves code documentation and maintainability by providing clear documentation for all generated functions.

Closes #59


mrscottyrose:

CoveMB:
Hey @mrscottyrose thanks for your MR ✨

Some considerations, when adding NatSpec we qwould probably want to align both content and formatting with what is found on the contract library [here is an example ](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Bridgeable.sol#L9)
We might not want to add comments to overrides.

Let's wait for @OpenZeppelin/contracts input for defining the content of the comments.

packages/core/solidity/src/account.ts
```diff
@@ -224,15 +224,6 @@ function overrideRawSignatureValidation(c: ContractBuilder, opts: AccountOptions
...
     c.addOverride({ name: 'AccountERC7579' }, signerFunctions._rawSignatureValidation);
-    c.setFunctionComments(
-      [
-        `// IMPORTANT: Make sure Signer${opts.signer} is most derived than AccountERC7579`,
-        `// in the inheritance chain (i.e. contract ... is AccountERC7579, ..., Signer${opts.signer})`,
-        '// to ensure the correct order of function resolution.',
-        '// AccountERC7579 returns false for `_rawSignatureValidation`',
-      ],
-      signerFunctions._rawSignatureValidation,
-    );
```
ernestognw, line 235: Any reason to remove this? It's an important consideration for anyone copy pastes the code

ernestognw (commented):
I like the idea at first. But I would prefer NatSpec documentation to be well thought and to provide useful information that's not available already in the documentation of the internal functions.

Since Wizard is a tool for developers, most of the information they'd need shows up on IDEs

Captura de pantalla 2025-05-15 a la(s) 2 56 01 p m

Concretely, I'd avoid using `@param` and `@return` tags since they bloat the result with things like

```solidity
/// @dev Executes a call to target with value
/// @param target The target that is going to be called
/// @param value The value to be sent
/// @param data The data to call with
function execute(address target, uint256 value, bytes data)
```

Instead, I think developers would benefit if we have more insightful descriptions in strategic functions (those not obvious). For example

```
/// @dev Performs a `call` to the `target` address. Consider override this function
/// to check if the target has code (_execute allows calling EOAs)
function execute(address target, uint256 value, bytes data) {
  _execute(target, value, data);
}
```

packages/core/solidity/src/account.ts
mrscottyrose: Sorry for that. I have reverted it back

mrscottyrose:
I've tried to resolve it according to the requirement could you please check and tell me if there's anything i could add or change?
Thank you.

packages/core/solidity/src/signer.ts
```diff
@@ -81,44 +81,53 @@ export const signerFunctions = {
...
+      args: [{ name: 'signer', type: 'address' }],
+      comments: [
+        '/// @dev Initializes the contract with an RSA signer. This function should be called during contract deployment to set up the signing mechanism.',
       ],
     },
     initializeMultisig: {
       kind: 'public' as const,
-      args: [
-        { name: 'signers', type: 'bytes[] memory' },
-        { name: 'threshold', type: 'uint256' },
```
CoveMB, line 103: `threshold` got removed, but should probably not

packages/core/solidity/src/signer.ts
```diff
@@ -81,44 +81,53 @@ export const signerFunctions = {
...
+      args: [{ name: 'signers', type: 'address[] memory' }],
+      comments: [
+        '/// @dev Initializes the contract with multiple signers. This function should be called during contract deployment to set up the multisig mechanism.',
       ],
     },
     initializeMultisigWeighted: {
       kind: 'public' as const,
       args: [
-        { name: 'signers', type: 'bytes[] memory' },
+        { name: 'signers', type: 'address[] memory' },
```
CoveMB, line 112: Let's revert all type changes has it they are out of scope for this MR

CoveMB:
Thanks for making changes, please revert any type / number of arguments changes from the MR

mrscottyrose:
@CoveMB what about now? I couldn't understand why the netlify checks failed.

packages/core/solidity/src/signer.ts
```diff
@@ -110,15 +122,21 @@ export const signerFunctions = {
...
+      comments: [
+        '/// @dev Initializes the contract with weighted signers. This function should be called during contract deployment to set up the weighted multisig mechanism.',
+      ],
     },
     _rawSignatureValidation: {
       kind: 'internal' as const,
       args: [
         { name: 'hash', type: 'bytes32' },
-        { name: 'signature', type: 'bytes calldata' },
+        { name: 'signature', type: 'calldata' },
```
CoveMB, line 133: Arguments change are out of scope of this MR

packages/core/solidity/src/erc721.ts
```diff
@@ -216,25 +216,34 @@ const functions = defineFunctions({
...
+      '/// @dev Returns the Uniform Resource Identifier (URI) for the specified token. This is used to store metadata for each token.',
+    ],
   },
 
   _baseURI: {
     kind: 'internal' as const,
     args: [],
     returns: ['string memory'],
-    mutability: 'pure' as const,
+    mutability: 'view' as const,
```
CoveMB, line 243: Should this mutability be changed? Probably out of scope

mrscottyrose:
@CoveMB are these two changes are enough to merge the mr?? Or I have to change other things also?

.changeset/six-yaks-accept.md
```diff
@@ -0,0 +1,13 @@
...
+
+Refine NatSpec documentation in generated contracts to improve developer experience:
+
+- Remove redundant parameter documentation where function signatures are self-explanatory
+- Add strategic documentation for non-obvious functions, focusing on implementation considerations
+- Preserve critical inheritance and function resolution comments in account contracts
+- Align documentation style with OpenZeppelin's contract library examples
+- Document override functions only when they add significant value
+
+This change follows the principle that documentation should provide insights beyond what's already visible in the code, while ensuring critical implementation details are preserved for developers.
```
ericglau, line 13: This should be less verbose.  It can be changed to something like:
```suggestion
Add NatSpec descriptions to generated contract functions.
```

mrscottyrose:
@CoveMB is there anything I need to change?

packages/core/solidity/src/erc20.ts
```diff
@@ -428,42 +421,100 @@ export const functions = defineFunctions({
...
 
   _checkTokenBridge: {
     kind: 'internal' as const,
     args: [{ name: 'caller', type: 'address' }],
+    comments: [
+      '/// @dev Validates if the caller is an authorized token bridge. This is used for cross-chain token transfers to ensure only trusted bridges can move tokens.',
+    ],
+  },
+
+  mint: {
```
CoveMB, line 510: Wondering why moving the functions orders here "mint"

packages/core/solidity/src/erc20.ts
```diff
@@ -428,42 +421,100 @@ export const functions = defineFunctions({
...
+    args: [
+      { name: 'account', type: 'address' },
+      { name: 'value', type: 'uint256' },
+    ],
+    comments: [
+      '/// @dev Destroys tokens from the specified account. This is an internal function that should be called from a public function with proper access control.',
+    ],
+  },
+
+  _spendAllowance: {
```
CoveMB, line 452: Wondering why adding this function?

packages/core/solidity/src/erc20.ts
```diff
@@ -428,42 +421,100 @@ export const functions = defineFunctions({
...
-      { name: 'amount', type: 'uint256' },
+      { name: 'account', type: 'address' },
+      { name: 'value', type: 'uint256' },
+    ],
+    comments: [
+      '/// @dev Creates new tokens and assigns them to the specified account. This is an internal function that should be called from a public function with proper access control.',
+    ],
+  },
+
+  _burn: {
```
CoveMB, line 441: Wondering why adding this function?

CoveMB:
@ernestognw could you validate/correct the comment's content?

packages/core/solidity/src/erc20.ts
```diff
@@ -428,7 +421,20 @@ export const functions = defineFunctions({
...
       { name: 'spender', type: 'address' },
       { name: 'value', type: 'uint256' },
-      { name: 'emitEvent', type: 'bool' },
+    ],
+    comments: [
+      '/// @dev Sets the allowance for a spender to spend tokens on behalf of an owner. This is an internal function that should be called from a public function with proper access control.',
+    ],
+  },
+
+  _mint: {
```
ericglau, line 430: This function isn't referenced directly, so it should be removed.

packages/core/solidity/src/erc721.ts
```diff
@@ -216,25 +216,34 @@ const functions = defineFunctions({
   _update: {
     kind: 'internal' as const,
     args: [
+      { name: 'from', type: 'address' },
```
ericglau, line 219: Args should not change.
```suggestion
```

packages/core/solidity/src/erc721.ts
```diff
@@ -216,25 +216,34 @@ const functions = defineFunctions({
...
+      '/// @dev Hook that is called during any token transfer. This includes minting and burning. Override this function to add custom logic for token transfers.',
+    ],
   },
 
   tokenURI: {
     kind: 'public' as const,
     args: [{ name: 'tokenId', type: 'uint256' }],
     returns: ['string memory'],
-    mutability: 'view' as const,
+    mutability: 'view',
```
ericglau, line 233: Why was `as const` removed?
```suggestion
    mutability: 'view' as const,
```

packages/core/solidity/src/erc721.ts
```diff
@@ -216,25 +216,34 @@ const functions = defineFunctions({
...
+      '/// @dev Returns the Uniform Resource Identifier (URI) for the specified token. This is used to store metadata for each token.',
+    ],
   },
 
   _baseURI: {
     kind: 'internal' as const,
     args: [],
     returns: ['string memory'],
-    mutability: 'pure' as const,
+    mutability: 'view',
```
ericglau, line 243: Mutability should not change.
```suggestion
    mutability: 'pure' as const,
```

packages/core/solidity/src/erc721.ts
```diff
@@ -243,10 +252,19 @@ const functions = defineFunctions({
...
       { name: 'value', type: 'uint128' },
     ],
+    comments: [
+      '/// @dev Increases the balance of an account. This is used internally to track token ownership and should not be called directly.',
+    ],
   },
 });
 
 function getMintFunction(incremental: boolean, uriStorage: boolean): BaseFunction {
+  /**
```
ericglau, line 262: Thanks for adding this doc comment, but it should be moved to above this `getMintFunction`.

packages/core/solidity/src/set-clock-mode.ts
```diff
@@ -26,7 +34,6 @@ export function setClockMode(c: ContractBuilder, parent: ReferencedContract, vot
     c.addOverride(parent, functions.clock);
     c.setFunctionBody(['return uint48(block.timestamp);'], functions.clock);
 
-    c.setFunctionComments(['// solhint-disable-next-line func-name-mixedcase'], functions.CLOCK_MODE);
```
ericglau, line 29: This solhint comment should still be included.  It could be appended to the newly added function comments.

packages/core/solidity/src/erc20.ts
```diff
@@ -400,16 +400,6 @@ function addSuperchainERC20(c: ContractBuilder) {
     functions._checkTokenBridge,
     'pure',
   );
-  c.setFunctionComments(
```
ericglau, line 403: This comment text is still relevant for this scenario.  It would be fine for it to overwrite the generic function comment for `_checkTokenBridge`

mrscottyrose:
@ericglau sorry I was messing everything again and again. Could you please review the new commit? If there are any changes let me know thanks

mrscottyrose:
@CoveMB can you check now? if everything is alright or not?

CoveMB:
It seem that [this comment](https://github.com/OpenZeppelin/contracts-wizard/pull/549/files#r2100434775) has not been addressed? (seem like the specialised comment is deleted instead of overriding the general comment)

mrscottyrose:
@CoveMB I ran the test and some of the tests fails cuz I'm in windows. What should I do? Will I push the generated test files?

CoveMB:
humm interesting could you share the issues here?

CoveMB:
Thanks for landing a hand @ObjectPlayer 🌾
Let us know if you can move forward or maybe you could also pull this commit from a fork of yours

ericglau:
Hi @mrscottyrose, thanks a lot for your efforts here.  However, from reviewing the proposed changes for these NatSpec descriptions, we’ve determined that these may not be beneficial for users, and instead they could make the code harder to read.

For function overrides due to multiple inheritance, of which there are many, these do not really benefit from NatSpec since users should just look at the imported (overridden) library functions' NatSpec.  For other functions, many have just one line implementations with potentially a modifier like `onlyOwner`.  Users should inspect these implementations rather than rely on NatSpec descriptions.  There are some cases where comments are useful, but these should be handled case-by-case, and comments are already included in some areas where needed (for example, in the Account tab).

For these reasons, we will close this PR and the original issue as won’t fix.

PR #554, opened 2025-05-19
Add onlyEntryPointOrSelf modifer to multisig

ernestognw:
Add missing modifers to multisigs


CoveMB (commented):
Is there a specific set of ok/fail tests that could be added?

ernestognw:
I think those are the snapshot tests, it's just difficult to get it at first when reviewing hundreds of combinations 😂

CoveMB (approved):
🌻

PR #558, opened 2025-05-26
Cairo add security contract field

### LINKED ISSUE #440: Cairo: Add security contact field
Adding a security contact field helps with vulnerability reporting for deployments with known source code. 

Wizard's Solidity convention is to use a custom NatSpec tag `/// @custom:security-contact security@example.com`

- [ ] Define what should be the convention for Cairo and other Rust-based languages
- [ ] Implement this for Cairo (and other languages where applicable)

CoveMB:
Fixes #440 (for Cairo and Cairo-alpha only)

- Add input UI component to capture security info
- Update info type for Cairo
- Add security info for AI agent
- Add contract documentation for Cairo contract builder
- Add print of contract documentation ( here security info) for Cairo
- Add test for printing documentation


ericglau (approved):
LGTM, thanks!
Note this should also be added to `cairo_alpha` (could be in this PR or a separate PR)

packages/core/cairo_alpha/src/utils/version.ts
```diff
@@ -14,4 +14,4 @@ export const scarbVersion = '2.11.4';
 /**
  * Semantic version string representing of the minimum compatible version of Contracts to display in output.
  */
-export const compatibleContractsSemver = '2.0.0-alpha.1';
+export const compatibleContractsSemver = '^2.0.0-alpha.1';
```
ericglau, line 17: This was intended to be pinned to a specific version, since the alpha may not be forwards-compatible.
```suggestion
export const compatibleContractsSemver = '2.0.0-alpha.1';
```

packages/core/cairo/src/print.ts
```diff
@@ -427,3 +429,7 @@ function printArgument(arg: Argument): string {
     return `${arg.name}`;
   }
 }
+
+function printDocumentationTags(tags: DocumentationTag[] = []): string[] {
```
ericglau, line 433: I think it would be more typesafe to not have a default value for the function parameter, so that the caller always needs to pass it in (even if it's just an empty array).

Same for cairo_alpha

immrsd (approved):
LGTM

ericnordelo (approved):
LGTM!

PR #561, opened 2025-05-30
Allow resizing controls

ernestognw:
### Description

In some contexts is convenient to change the width of the controls.

https://github.com/user-attachments/assets/ce019b23-d946-4c08-abd3-6874695441cc

### DIFF
--- a/packages/ui/src/cairo/App.svelte
+++ b/packages/ui/src/cairo/App.svelte
@@ -179,7 +179,7 @@
   </div>
 
   <div class="flex flex-row grow">
-    <div class="controls rounded-l-3xl w-72 flex flex-col shrink-0 justify-between h-[calc(100vh-84px)] overflow-auto">
+    <div class="controls rounded-l-3xl min-w-72 w-72 max-w-[calc(100vw-420px)] flex flex-col shrink-0 justify-between h-[calc(100vh-84px)] overflow-auto resize-x">
       <div class:hidden={tab !== 'ERC20'}>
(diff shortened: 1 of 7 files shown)


ericglau (commented):
Nice! Just a few comments:
- If you expand the nav almost all of the way to the right:
  1. The "Deploy" button partly gets covered.
  2. If you click "Deploy", the arrows to close the Deploy modal get hidden.  This may need some adjusting of the z-indexes, such as [here](https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/ui/src/solidity/App.svelte#L378).

https://github.com/user-attachments/assets/a2b40ea8-20e7-4b7d-b9d6-c238a0eda70f

- Can you apply this change to the other languages under `packages/ui/src/*/App.svelte`?

ernestognw:
Thanks for reviewing @ericglau!

> If you expand the nav almost all of the way to the right

Right! I didn't consider much of the potential UI issues on the borders. I also think it makes sense to restrict to a minimum width. So I set it to `min-w-72` and for the maximum I took the fixed width of the Defender sidebar (360px) plus roughly 40 px for the close button (It's weird to measure, it's kinda floating).

The constrained sizes should make it better now

> The space between expand/collapse arrows and the question mark isn't constant.

Also pretty good catch! I adjusted and removed an unused `align` property in the `HelpTooltip` component. Should be fine now

> Can you apply this change to the other languages under packages/ui/src/*/App.svelte?

Done!

ericglau (approved):
LGTM, thanks!

PR #562, opened 2025-06-03
Download stellar scaffold

CoveMB:
## Download stellar scaffold project 

The idea is to set up the project with minimal changes from the latest version of scaffold project changes (potential changes in configuration files) and only make changes to what we know need changes and can be infered from the Wizard's contract.

- Add a button in the UI to download a [Scaffold project](https://github.com/AhaLabs/scaffold-stellar/tree/main) from Wizard Stellar page
- The button will download a zip folder tha will include the following
    -  A `README.md` file explaining how to set up the project, run test and install required dependencies
    - A `contracts` folder containing 
        - a folder for the `contract` code _named accordingly to the type of contract downloaded_ containing
            - A `Cargo.toml` file with appropriate dependencies _dependencies are computed depending on the options selected in the Wizard UI_ redirecting to the workspace dependencies
            -  A `src` folder containing
                -  A `contract.rs` file containing the code displayed in Wizard _minus the crate level attribute that is moved to `lib.rs` file_
                - A `test.rs` file containing a simple test testing for the contract name _available for both upgradable and non-upgradable contracts_
                -  A `lib.rs` file that include both files for cargo's awareness
    - A `setup.sh` file that will set up the project as following
        - Check for required dependencies to run Scaffold project
            -  git
            - cargo
            - stellar
            - docker
            - node
        - Save the README.md as WIZARD-README.md
        - Download the `scaffold` initial project _and delete the contracts folder as well as moving all file one level up in the directory tree_
        - Set up the `environments.toml` file _by removing previously configured contracts, adding the current contracts from Wizard and inserting a comment in the constructor of the contracts if the contracts is upgradable, _
        - Update the workspace `Cargo.toml` file with the dependencies version of Wizard's
        - Build the contract
        - Install NPM dependencies
        - Initialize Git

### Note
Will add tests once the main idea and set-up flow is validated

### DIFF
--- a/packages/core/stellar/src/zip-scaffold.test.ts
+++ b/packages/core/stellar/src/zip-scaffold.test.ts
@@ -125,432 +125,432 @@ test('contractOptionsToScaffoldContractName converts PascalCase to snake_case', 
   t.is(contractOptionsToScaffoldContractName('MyCustomKind'), 'my_custom_kind');
 });
 
-test('fungible simple', async t => {
-  const opts: GenericOptions = {
-    kind: 'Fungible',
-    name: 'MyToken',
-    symbol: 'MTK',
-    premint: undefined,
-    burnable: false,
-    mintable: false,
-    pausable: false,
-    upgradeable: false,
-  };
-  const c = buildFungible(opts);
-  await runTest(t, c, opts);
-});
+// test('fungible simple', async t => {
+//   const opts: GenericOptions = {
+//     kind: 'Fungible',
+//     name: 'MyToken',
+//     symbol: 'MTK',
+//     premint: undefined,
+//     burnable: false,
+//     mintable: false,
+//     pausable: false,
+//     upgradeable: false,
+//   };
+//   const c = buildFungible(opts);
+//   await runTest(t, c, opts);
+// });
 
-test('fungible full', async t => {
-  const opts: GenericOptions = {
-    kind: 'Fungible',
-    name: 'MyToken',
-    symbol: 'MTK',
-    premint: '2000',
-    burnable: true,
-    mintable: true,
-    pausable: true,
-    upgradeable: false,
-  };
-  const c = buildFungible(opts);
-  await runTest(t, c, opts);
-});
+// test('fungible full', async t => {
+//   const opts: GenericOptions = {
+//     kind: 'Fungible',
+//     name: 'MyToken',
+//     symbol: 'MTK',
+//     premint: '2000',
+//     burnable: true,
+//     mintable: true,
+//     pausable: true,
+//     upgradeable: false,
+//   };
+//   const c = buildFungible(opts);
+//   await runTest(t, c, opts);
+// });
 
-test('fungible burnable', async t => {
-  const opts: GenericOptions = {
-    kind: 'Fungible',
-    name: 'MyToken',
-    symbol: 'MTK',
-    premint: '2000',
-    burnable: true,
-    mintable: false,
-    pausable: false,
-    upgradeable: false,
-  };
-  const c = buildFungible(opts);
-  await runTest(t, c, opts);
-});
+// test('fungible burnable', async t => {
+//   const opts: GenericOptions = {
+//     kind: 'Fungible',
+//     name: 'MyToken',
+//     symbol: 'MTK',
+//     premint: '2000',
+//     burnable: true,
+//     mintable: false,
+//     pausable: false,
+//     upgradeable: false,
+//   };
+//   const c = buildFungible(opts);
+//   await runTest(t, c, opts);
+// });
 
-test('fungible mintable', async t => {
-  const opts: GenericOptions = {
-    kind: 'Fungible',
-    name: 'MyToken',
-    symbol: 'MTK',
-    premint: '2000',
-    burnable: false,
-    mintable: true,
-    pausable: false,
-    upgradeable: false,
-  };
-  const c = buildFungible(opts);
-  await runTest(t, c, opts);
-});
+// test('fungible mintable', async t => {
+//   const opts: GenericOptions = {
+//     kind: 'Fungible',
+//     name: 'MyToken',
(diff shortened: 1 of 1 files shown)


packages/core/stellar/src/zip-scaffold.ts
```diff
@@ -0,0 +1,346 @@
...
+- See [Scaffold Cli installation guide](https://github.com/AhaLabs/scaffold-stellar?tab=readme-ov-file#quick-start).
+- See [Docker installation guide](https://docs.docker.com/engine/install/).
+- See [Node installation guide](https://nodejs.org/en/download).
+
+## Initializing the project
+
+\`\`\`
+bash setup.sh
+\`\`\`
+${hasTodosToResolve(c) ? '\n## Resolve any TODOs \n\nSearch for any TODO comments in the project and resolve them (search for TODO with your code editor)\n' : ''}
```
ericglau, line 304:
```suggestion
${hasTodosToResolve(c) ? '\n## Resolve any TODOs \n\nSearch for any TODO comments in the project and resolve them (search for TODO with your code editor).\n' : ''}
```

packages/core/stellar/src/zip-scaffold.ts
```diff
@@ -0,0 +1,346 @@
...
+
+use soroban_sdk::{ ${opts.upgradeable ? 'testutils::Address as _, Address, ' : ''}Env, String };
+
+use crate::contract::{ ${c.name}, ${c.name}Client };
+
+#[test]
+fn initial_state() {
+    let env = Env::default();
+
+    let contract_addr = env.register(${c.name}, (${opts.upgradeable ? 'Address::generate(&env),' : ''}));
```
ericglau, line 42: The second parameter (for the constructor args) should be based on the number of actual constructor args (perhaps using `getAddressArgs` from above, assuming all args are addresses).

For example, if the user chooses fungible with a premint amount and enables Mintable, the constructor is `__constructor(e: &Env, recipient: Address, owner: Address)` so there needs to be two addresses passed in.

ericglau (commented):
Looks great! Just a few comments.

packages/core/stellar/src/zip-scaffold.ts
```diff
@@ -0,0 +1,346 @@
...
+
+    assert_eq!(client.name(), String::from_str(&env, "${c.name}"));
+}
+    
+// Add more tests bellow
+`;
+
+function contractCargo(opts: GenericOptions, scaffoldContractName: string) {
+  return `[package]
+name = "${scaffoldContractName}_contract"
```
brozorec, line 53: this should be better kebab-case
```suggestion
name = "${scaffoldContractName.replace(/_/, '-')}-contract"
```

packages/core/stellar/src/zip-scaffold.ts
```diff
@@ -0,0 +1,346 @@
...
+edition.workspace = true
+license.workspace = true
+repository = "https://github.com/OpenZeppelin/stellar-contracts"
```
brozorec, line 56: this should be user's own repositiory, I suggest we completely remove this attribute from "Cargo.toml"

packages/core/stellar/src/zip-scaffold.ts
```diff
@@ -0,0 +1,346 @@
...
+check_is_installed stellar "Scaffold"
+check_is_installed docker "Docker"
+check_is_installed node "Node"
+
+
+if ! [ -f "environments.toml" ]
+then
+  echo "🏗️ Building Scaffold project"
+
+  mv README.md WIZARD-README.md
```
brozorec, line 262: When I run the setup, I was surprised to see the initial readme changed. Should we have it as "WIZARD-README.md" from the beginning?

packages/core/stellar/src/zip-scaffold.ts
```diff
@@ -0,0 +1,346 @@
...
+  }, '');
+
+const getDependenciesToAdd = (opts: GenericOptions): (keyof typeof dependencies)[] => {
+  const dependenciesToAdd: (keyof typeof dependencies)[] = [];
+
+  if (opts.kind === 'Fungible') dependenciesToAdd.push('fungible');
+  if (opts.kind === 'NonFungible') dependenciesToAdd.push('nonFungible');
+  if (opts.pausable) dependenciesToAdd.push('pausable');
+  if (opts.upgradeable) dependenciesToAdd.push('upgradable');
+
```
brozorec, line 26: In most of the cases we need to add another dependency "stellar-default-impl-macro". Instead of adding conditions on when to add it, I suggest we just include it by default.

brozorec (changes requested):
Thank you, great job 👍

ericglau (approved):

.changeset/quick-views-live.md
```diff
@@ -0,0 +1,2 @@
+---
```
ericglau, line 1: Instead of using an empty changeset, we should just tag the PR with label `ignore-changeset`

packages/core/stellar/src/zip-scaffold.test.ts
```diff
@@ -0,0 +1,588 @@
...
+
+async function runTest(t: ExecutionContext<Context>, c: Contract, opts: GenericOptions) {
+  t.timeout(300_000);
+
+  const zip = await zipScaffold(c, opts);
+
+  assertLayout(t, zip, opts);
+  await extractPackage(t, zip);
+  await runProjectSetUp(t);
+  await runContractTest(t);
```
ericglau, line 113: The running of the project's scripts/tests are a great addition, but now the overall stellar checks on GitHub take 57 minutes.

This works for now, but we can consider one of the below as a later follow-up:
- Once we have compilation tests for Wizard's Stellar contracts in general, we can simplify this to only do the project setup/tests for a small selection of the test cases below.
- Or, can the testcases below be parallelized?
- Or, should these scaffold tests be run in a separate GitHub check?

PR #564, opened 2025-06-06
Stylus add security contract field

CoveMB:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/440 (for Stylus only)

Add input UI component to capture security info
Update info type for Stylus
Add security info for AI agent
Add contract documentation for Stylus contract builder
Add print of contract documentation (here security info) for Stylus
Add test for printing documentation


packages/core/stylus/src/contract.test.ts.md
```diff
@@ -167,3 +167,26 @@ Generated by [AVA](https://avajs.dev).
...
+
+    `// SPDX-License-Identifier: MIT␊
+    // Compatible with OpenZeppelin Contracts for Stylus ^0.2.0-alpha.4␊
+    ␊
+    #![cfg_attr(not(any(test, feature = "export-abi")), no_main)]␊
+    extern crate alloc;␊
+    ␊
+    use stylus_sdk::prelude::*;␊
+    ␊
+    /// @custom:security-contact security@example.com␊
```
CoveMB, line 183: @0xNeshi @bidzyyys  Would this be the appropriate placement for the security contact comment (and possibly other contract level comments)?

0xNeshi: It's fine I guess, but a more idiomatic (Rust) way would be to put them at the module docs level `//!`, right after the SPDX comment:
```rust
// Compatible with OpenZeppelin Contracts for Stylus ^0.2.0-alpha.4

//! Security contact: security@example.com

#![cfg_attr(...
...
```

A more descriptive comment would be:
```rust
//! # Security
//! 
//! For security issues, please contact: security@example.com
```

CoveMB: Noted, I agree this is clearer and more descriptive, made the changes (note that it then differ more than the custom [Natspec like standarts](https://github.com/OpenZeppelin/contracts-wizard/issues/440))

packages/core/stylus/package.json
```diff
@@ -28,4 +29,4 @@
     "typescript": "^5.0.0",
     "semver": "^7.6.0"
   }
-}
+}
```
0xNeshi, line 32:
```suggestion
}

```
nit: missing newline char

packages/core/stylus/src/contract.test.ts.md
```diff
@@ -167,3 +167,50 @@ Generated by [AVA](https://avajs.dev).
...
+    ␊
+    use stylus_sdk::prelude::*;␊
+    ␊
+    #[entrypoint]␊
+    #[storage]␊
+    struct Foo {}␊
+    ␊
+    #[public]␊
+    #[inherit]␊
+    impl Foo {}␊
```
0xNeshi, line 190: I think we should recommend devs use idiomatic Rust docs as opposed to NatSpec, just as we do in our library, wdyt @bidzyyys @qalisander?

I.e. instead of:
```rust
// @custom:documentation some documentation
```
should use:
```rust
//! Some custom documentation
```

@CoveMB is there a benefit to using NatSpec with Stylus?

0xNeshi (approved):

PR #565, opened 2025-06-06
Cairo alpha test snapshot update

CoveMB:
Seem like Cairo alpha test snapshot did not get updated after fixing version


ericglau (approved):

packages/core/cairo_alpha/src/account.test.ts.md
```diff
@@ -9,7 +9,7 @@ Generated by [AVA](https://avajs.dev).
 > Snapshot 1
 
     `// SPDX-License-Identifier: MIT␊
-    // Compatible with OpenZeppelin Contracts for Cairo ^2.0.0-alpha.1␊
```
ericnordelo, line 12: Is removing the caret a change we only introduced for alphas, right?

CoveMB: Exact

PR #566, opened 2025-06-09
Stellar: fix default_impl and ContractOverrides

brozorec:
This PR:
- renames `defaultimpl` to `default_impl` in Non-Fungible
- adds `ContractOverrides` to imports in Enumerable

### DIFF
--- a/.changeset/pink-planes-talk.md
+++ b/.changeset/pink-planes-talk.md
@@ -2,5 +2,4 @@
 '@openzeppelin/wizard-stellar': patch
 ---
 
-- Fix missing `ContractOverrides` import.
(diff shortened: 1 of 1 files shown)


CoveMB (approved):
Sounds good, we can consider adding compilation tests so we can test the compilation of all different combination.

I started [here](https://github.com/OpenZeppelin/contracts-wizard/pull/562/files#diff-aaf227fd5d58ae5346babec6437cda89bd87bfd6730456927628da9cf7cdd328R1) we could replace running test with simply running `cargo build` and also catch warnings

ericglau (approved):

PR #574, opened 2025-06-18
Merge Cairo alpha to stable v2.0.0

ericglau:
- Merges the changes from Cairo alpha into stable for v2.0.0
- If Cairo alpha and stable versions are the same, hides the version dropdown from the UI


ericnordelo (approved):
LGTM!

packages/core/cairo/src/test.ts
```diff
@@ -54,7 +54,7 @@ async function testGenerate(params: {
   royaltyInfo?: RoyaltyInfoSubset;
 }) {
   const { ctx, kind, royaltyInfo } = params;
-  const generatedSourcesPath = path.join(os.tmpdir(), 'oz-wizard-cairo');
+  const generatedSourcesPath = path.join(os.tmpdir(), 'oz-wizard-cairo-alpha');
```
CoveMB, line 57: We need this "alpha" in folder name?

ericglau: good catch, removed

PR #576, opened 2025-06-19
Add account factory

Amxx:


packages/core/solidity/src/account.ts
```diff
@@ -244,53 +249,105 @@ function overrideRawSignatureValidation(c: ContractBuilder, opts: AccountOptions
...
+        `Clones.cloneDeterministic(address(implementation), effectiveSalt);`,
+        `${accountName}(instance).${accountInitializer.name}(${accountInitializer.args.map(arg => arg.name).join(', ')});`,
       ],
-    },
-  }),
-};
+      `}`,
+      `return instance;`,
+    ]).split('\n'),
+    { name: 'create', kind: 'public' as const, args, returns: [ 'address' ] },
```
ericglau, line 288: The function name is `create` but the cloning is using create2.  These are unrelated but should we make it clearer that this deploys accounts using create2?

ericglau (commented):
Overall this looks like a good approach.

PR #577, opened 2025-06-20
Update account generation to use code from 5.4.0-rc.1

Amxx:
Upgradeable version is a bit challenging because only part of the parent contracts must be used in their upgradeable version.

Only contract that should be used from the upgradeable packages are:
```
draft-AccountERC7579Hooked.sol
draft-AccountERC7579HookedUpgradeable.sol
MultiSignerERC7913Upgradeable.sol
MultiSignerERC7913WeightedUpgradeable.sol
SignerECDSAUpgradeable.sol
SignerERC7913Upgradeable.sol
SignerP256Upgradeable.sol
SignerRSAUpgradeable.sol
```

while `EIP712`, and `ERC7739` have an upgradeable version, they should not be used here. Using the vanila version of `EIP712` so that the domain doesn't have to be initialized, and is shared accross all clones/proxies.


packages/core/solidity/src/account.ts
```diff
@@ -85,36 +86,32 @@ function addParents(c: ContractBuilder, opts: AccountOptions): void {
...
-          functions.isValidSignature,
-        );
-      }
-      break;
+  if (opts.signatureValidation === 'ERC7739') {
+    c.addParent({
+      name: 'ERC7739',
+      path: '@openzeppelin/contracts/utils/cryptography/signers/draft-ERC7739.sol',
+    });
+  } else if (opts.signatureValidation === 'ERC1271' && !opts.signatureValidation) {
```
ericglau, line 102: This condition is not possible.

packages/core/solidity/src/signer.ts
```diff
@@ -1,110 +1,124 @@
...
-      c.addFunctionCode(`_setThreshold(${fn.args[2]!.name});`, fn);
+    case false: {
+      return;
+    }
+    case 'ERC7702': {
+      // Add locking constructor
+      c.addNatspecTag('@custom:oz-upgrades-unsafe-allow', 'constructor');
+      c.addConstructorCode(`_disableInitializers();`);
+
+      // todo: error if upgradeable is set ?
```
ericglau, line 19: If this should be an error, it can throw `OptionsError` and the UI can catch and display it.  See https://github.com/OpenZeppelin/contracts-wizard/blob/4b86b076214b6aa9b62e472b431d5d2ffdd96ffb/packages/core/solidity/src/governor.ts#L198 and https://github.com/OpenZeppelin/contracts-wizard/blob/4b86b076214b6aa9b62e472b431d5d2ffdd96ffb/packages/ui/src/solidity/GovernorControls.svelte#L90 for example.

packages/core/solidity/package.json
```diff
@@ -23,8 +23,8 @@
   },
   "devDependencies": {
     "@openzeppelin/community-contracts": "https://github.com/OpenZeppelin/openzeppelin-community-contracts",
-    "@openzeppelin/contracts": "^5.3.0",
-    "@openzeppelin/contracts-upgradeable": "^5.3.0",
+    "@openzeppelin/contracts": "^5.4.0-rc.1",
+    "@openzeppelin/contracts-upgradeable": "^5.4.0-rc.1",
```
ericglau, line 27: These should be bumped to 5.4.0 when it is released, before merging this PR.

ericglau (changes requested):
Thanks for the PR.  Here is a checklist of some remaining items that are needed:
- [ ] The minimum version should be updated to `^5.4.0` in `packages/core/solidity/src/utils/version.ts`
- [ ] Please add a changeset, mentioning that this is a potentially breaking change because it requires Contracts v5.4.0.
- [ ] An Upgradeability section should be added to the Account tab in the UI, in `packages/ui/src/solidity/AccountControls.svelte`
- [ ] Should the Experimental status be removed from the Account tab in the UI?  If so, the following need to be updated:
  - https://github.com/OpenZeppelin/contracts-wizard/blob/df1bdffdd93435d99b80da391b5550a98b03e38d/packages/ui/src/solidity/App.svelte#L247
  - https://github.com/OpenZeppelin/contracts-wizard/blob/df1bdffdd93435d99b80da391b5550a98b03e38d/packages/ui/src/solidity/App.svelte#L261
  - https://github.com/OpenZeppelin/contracts-wizard/blob/df1bdffdd93435d99b80da391b5550a98b03e38d/packages/ui/src/solidity/AccountControls.svelte#L27-L30
  - The note in https://github.com/OpenZeppelin/contracts-wizard/blob/update/account-from-vanilla-5.4.0-rc/packages/core/solidity/README.md#contract-types
- [ ] Evaluate or test whether this works with Remix, downloaded Hardhat package, or downloaded Foundry package.  If it should work, update the conditions in https://github.com/OpenZeppelin/contracts-wizard/blob/df1bdffdd93435d99b80da391b5550a98b03e38d/packages/ui/src/solidity/App.svelte#L150-L167 
- [ ] Schemas/definitions for function calling from AI assistants/MCP should be updated to allow upgradeability for Account.  This includes:
  - https://github.com/OpenZeppelin/contracts-wizard/blob/df1bdffdd93435d99b80da391b5550a98b03e38d/packages/ui/api/ai-assistant/function-definitions/solidity.ts#L235-L239
  - `packages/mcp/src/solidity/schemas.ts`

ericglau:
Replaced by https://github.com/OpenZeppelin/contracts-wizard/pull/609

PR #584, opened 2025-06-26
Refactor Stylus to the new Contract Format (post- Stylus SDK v0.9.0)

0xNeshi:
Closes https://github.com/OpenZeppelin/rust-contracts-stylus/issues/693


ericglau:
Can you revert the changelog and package.json version updates, and [add a changeset](https://github.com/OpenZeppelin/contracts-wizard/blob/master/CONTRIBUTING.md#adding-changesets) for `@openzeppelin/wizard-stylus` with type `minor`?  This will handle the bump to `0.2.0` automatically during the release process.

packages/core/stylus/src/contract.ts
```diff
@@ -1,24 +1,34 @@
 import { escapeString, toIdentifier } from './utils/convert-strings';
+import { copy } from './utils/copy';
 
 type Name = {
   identifier: string;
   stringLiteral: string;
 };
 
+export type TraitName = string;
```
CoveMB, line 9: Would it be possible to have a stronger type than `string` (like `IPausable | ..`?

packages/core/stylus/src/print.ts
```diff
@@ -123,111 +171,276 @@ function splitLongLineInner(line: string): Lines[] {
...
 
-  const sortedGroups = Object.entries(grouped).sort((a, b) => a[0].localeCompare(b[0]));
-  return sortedGroups;
+/**
+ * Extract errors from the contract and determine the error type.
+ * @returns {ErrorPrintData | undefined} if errors exist, undefined otherwise.
+ * For single unwrapped errors, returns inherited type.
+ * For multiple errors, returns enum type with grouped errors.
+ */
+function extractErrors(contract: Contract): ErrorPrintData | undefined {
```
CoveMB, line 205: nits: we could add small unit tests for those print functions

packages/core/stylus/src/contract.test.ts
CoveMB, line 1: wondering if all path are test covered? for example from wrapped trait and error `mergeWrappedErrors` logics

packages/core/stylus/src/utils/define-functions.ts
```diff
@@ -1,9 +1,7 @@
 import type { BaseFunction } from '../contract';
 
-type ImplicitNameFunction = Omit<BaseFunction, 'name'>;
+export function defineFunctions<F extends string>(fns: Record<F, BaseFunction>): Record<F, BaseFunction>;
```
ericglau, line 3: This requires the `name` field to be defined, which in most cases is the same as the key.

Not a must for this PR, but I'd suggest reverting this to keep the code shorter when defining functions in erc20.ts etc.

Alternatively, if you have a need to use different names than the keys, you could use implicit names but make them overridable.   See [Stellar as an example.](https://github.com/OpenZeppelin/contracts-wizard/blob/e8ae9d460d789e1002569cbb0454c2b77633f349/packages/core/stellar/src/utils/define-functions.ts).

ericglau (approved):
LGTM, thanks! Just one additional suggestion.

packages/core/stylus/src/add-pausable.ts
```diff
@@ -1,37 +1,33 @@
...
-    },
-    modulePath: 'openzeppelin_stylus::utils',
-  };
-
-  c.addFunction(pausableTrait, functions.pause);
-  c.addFunction(pausableTrait, functions.unpause);
-
-  requireAccessControl(c, pausableTrait, functions.pause, access, 'PAUSER', 'pauser');
-  requireAccessControl(c, pausableTrait, functions.unpause, access, 'PAUSER', 'pauser');
+  // requireAccessControl(c, pausableTrait, functions.pause, access, 'PAUSER', 'pauser');
```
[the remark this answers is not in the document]
0xNeshi: Leftover from when we thought `Pausable` would be added sooner, will remove :+1:

packages/core/stylus/src/contract.ts
0xNeshi: like it, implementing it

packages/core/stylus/src/print.ts
0xNeshi: The amount of effort would outweigh any potential benefits imo. The fact that all possible contracts generated in tests are valid is evidence enough that this function works as intended. 

The function would warrant more tests if its output was unpredictable, e.g. if its input was provided by users.

packages/core/stylus/src/contract.test.ts
0xNeshi: @CoveMB @ericglau I think these tests are very useful for validating how the contract generation logic behaves for [unpredictable (user) input](https://github.com/OpenZeppelin/contracts-wizard/pull/584#discussion_r2192184043), but if all possible contracts are validly generated in tests, what purpose do these generic tests serve?

Just checking whether it's worth putting effort into these `contract.test.ts` tests.

packages/core/stylus/src/utils/define-functions.ts
0xNeshi: Despite the redundancy, it might make code more concise by keeping this update.

The reason I left the "duplicate" `name` field is twofold:
1. Makes it much easier to convert this type-safe `functions` object into a `functions` array when defining the trait object (see [erc20Trait](https://github.com/OpenZeppelin/contracts-wizard/pull/584/files#diff-17e6754c7f8296c839f4f1d7deee044d7db13913e50fe94affced96846c52711R233)); if this `name` field was omitted (as it was before), each of these object->array conversions would require a verbose mapping to the correct format.
2. Makes it more straightforward to add/update functions manually, e.g. when adding [Erc721Enumerable](https://github.com/OpenZeppelin/contracts-wizard/pull/584/files#diff-004e40be3af42731e6cd56ef522b9f35589f78d0be4a4356bd138d942ed109fbR94-R117), otherwise a verbose mapping would be necessary.

bidzyyys (approved):
Looks good, left some minor comments.

packages/core/stylus/src/utils/define-functions.ts
0xNeshi: @ericglau pinging you in case you missed this, are you OK with keeping `name` as-is?

packages/core/stylus/src/contract.test.ts
0xNeshi: @CoveMB @ericglau pinging you in case you missed this

packages/core/stylus/src/erc20.test.ts.md
```diff
@@ -126,38 +233,74 @@ Generated by [AVA](https://avajs.dev).
...
-        #[borrow]␊
         erc20: Erc20,␊
-        #[borrow]␊
         flash_mint: Erc20FlashMint,␊
     }␊
     ␊
     #[public]␊
-    #[inherit(Erc20)]␊
-    impl MyToken {␊
+    #[implements(IErc20<Error = flash_mint::Error>, IErc3156FlashLender<Error = flash_mint::Error>)]␊
```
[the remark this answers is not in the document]
bidzyyys: Oh, okay, fair enough!

packages/core/stylus/src/erc1155.test.ts.md
```diff
@@ -9,28 +9,68 @@ Generated by [AVA](https://avajs.dev).
...
+    #[public]␊
+    impl IErc1155 for MyToken {␊
+        type Error = erc1155::Error;␊
+    ␊
+        fn balance_of(&self, account: Address, id: U256) -> U256 {␊
+            self.erc1155.balance_of(account, id)␊
+        }␊
+    ␊
+        fn balance_of_batch(&self, accounts: Vec<Address>, ids: Vec<U256>) -> Result<Vec<U256>, Self::Error> {␊
+            Ok(self.erc1155.balance_of_batch(accounts, ids)?)␊
```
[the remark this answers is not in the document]
bidzyyys: I see, it's a valid point!

PR #595, opened 2025-07-04
Add Superchain interop message passing

ericglau:
Remaining items:
- [ ] Add unit tests
- [ ] Add snapshot tests
- [ ] Update AI assistant definitions and MCP schemas
- [ ] Add Optimism contracts for compilation tests
- [ ] Add Optimism contracts to Download Hardhat and Foundry packages
- [ ] Test an example contract, non-upgradeable
- [ ] Test an example contract, upgradeable


gonzaotc:
Solid!, probably you're already aware but the ai assistant seems to be generating crosschain messaging code very well but the changes it doesn't gets reflected on the UI; while the code appears, the checkboxes kept unchecked as if nothing happened.

packages/core/solidity/src/zip-foundry.ts
gonzaotc, line 1: I'm curious, why are we using Soldeer for Optimism contracts but Foundry git submodules for the OpenZeppelin dependencies on the Foundry zip downloads?

Wouldn't it be easier and cleaner for developers to get a single package managing method instead of two?, or there is a hard requirement that forces us to use both?

gonzaotc (approved):
LGTM!. Tested in supersim and succeeded. Just added one non-blocking comment, great work Eric!.

packages/core/solidity/src/zip-foundry.ts
ericglau: For Optimism contracts, we have 4 options:
1. Not import them, and just hardcode the messenger contract's predeploy address and interface's function calls.
2. Use git submodules with forge install `ethereum-optimism/optimism`
3. Use NPM package `@eth-optimism/contracts-bedrock`
4. Use Soldeer package https://soldeer.xyz/project/@eth-optimism-contracts-bedrock which is a Soldeer-maintained mirror of the above NPM package.

For Foundry specifically, here are the drawbacks of each option:
1. This works, but it seems better practice to import the address and interface since they are available.
2. This requires downloading the entire repo which is over 500 MB and seems excessive for what we need.
3. This requires NPM to be installed, and is not Foundry native.
4. Depends on Soldeer and its repository, but this is natively supported in Foundry.

Option 4 seems the most appropriate for Optimism contracts in Foundry in this case.

For OpenZeppelin Contracts dependencies, current recommendations in https://github.com/OpenZeppelin/openzeppelin-contracts are to use forge install.  We can consider being consistent in Wizard and use Soldeer for this as well, but we should review the recommended patterns when doing so, and this should occur in a separate PR.

packages/ui/src/solidity/superchain-tooltip.ts
```diff
@@ -1,12 +1,22 @@
...
+  content: '<strong>Important:</strong> Only available on chains in the Superchain.',
   trigger: 'manual',
   placement: 'bottom',
   maxWidth: '22em',
   allowHTML: true,
   interactive: true,
 };
+
+const ERC20_REQUIREMENTS =
+  ' Requires deploying your contract to the same address on every chain in the Superchain. <a class="light-link" href="https://docs.optimism.io/stack/interop/superchain-erc20#requirements" target="_blank" rel="noopener noreferrer">Read more.</a>';
```
CoveMB, line 14: super nit: should we remove the space and have + " " + ERC20Req.. bellow?

packages/core/solidity/src/zip-foundry.ts
```diff
@@ -176,6 +177,11 @@ const script = (c: Contract, opts?: GenericOptions) => {
   }
 };
 
+const OPTIMISM_NPM_PACKAGE = '@eth-optimism/contracts-bedrock';
+const OPTIMISM_SOLDEER_PACKAGE = '@eth-optimism-contracts-bedrock';
+const importsOptimism = (c: Contract) => c.imports.some(i => i.path.startsWith(OPTIMISM_NPM_PACKAGE));
```
CoveMB, line 182: super nit: maybe "shouldImportOptimism" as it result in a boolean?

CoveMB (approved):
Nice work 🔥
Great test coverage
I am wondering in the Custom contract it seem Optimism option add both emitter and receiver functions is it standar or does contract usually implement on or the other?

ericglau:
@CoveMB 
> I am wondering in the Custom contract it seem Optimism option add both emitter and receiver functions is it standar or does contract usually implement on or the other?

In this example, we include both functions.  If only one function is included, the emitter and/or receiver would need to take or track other addresses, which would be a more complex example.  That can be considered as an enhancement if there is a need for it.

PR #609, opened 2025-07-25
Update migrated imports from community to vanilla contracts

### LINKED ISSUE #618: Use Account and signers from Contracts 5.4.0 instead of Community Contracts
Started in:
https://github.com/OpenZeppelin/contracts-wizard/pull/609
https://github.com/OpenZeppelin/contracts-wizard/pull/577

ernestognw:
Fixes #618

### DIFF
--- a/.changeset/kind-poems-pull.md
+++ b/.changeset/kind-poems-pull.md
@@ -1,0 +1,5 @@
+---
+'@openzeppelin/wizard-common': patch
+---
+
+Update Solidity Account prompt


packages/core/solidity/src/signer.ts
```diff
@@ -1,112 +1,114 @@
...
+    case 'Multisig':
+    case 'MultisigWeighted': {
+      if (upgradeable) {
+        c.addParent({
+          name: 'Initializable',
+          path: '@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol',
+        });
+        // Add locking constructor
+        c.addNatspecTag('@custom:oz-upgrades-unsafe-allow', 'constructor');
+        c.addConstructorCode(`_disableInitializers();`);
```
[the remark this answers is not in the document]
ernestognw: I think it's not worth it. In this case, there is a `setUpgradeableAccount` function that disables upgradeability, this is with the purpose of excluding the upgradeable versions of EIP712 and ERC7739. Since `upgradeable == false`, then the constructor will not set, this is why we need to manually add `_disableInitializers()` again.

ernestognw:
> Thanks @ernestognw! I added the upgradeable option for the account compilation tests, which revealed some compile errors.

Thanks for pointing out @ericglau. It seems that the Initializable error is still happening since the one imported when the upgradeable checkbox is set is the one in vanilla contracts: https://github.com/OpenZeppelin/contracts-wizard/blob/7bf5e16c901e2f62598e2394c446b804d5c31166/packages/core/solidity/src/set-upgradeable.ts#L21-L24

I'm not sure what to do, maybe changing the line in set-upgradeable will do the job but I remember we wanted to recommend using the `Initializable` from vanilla.

packages/core/solidity/src/signer.ts
```diff
@@ -1,112 +1,114 @@
...
+
+        const fn = { name: 'initialize', kind: 'public' as const, args: signerArgs[signer] };
+        // c.addModifier('initializer', fn);
+        c.addFunctionCode(`__${signers[signer].name}_init(${signerArgs[signer].map(arg => arg.name).join(', ')});`, fn);
+        c.addParent({
+          name: `${signers[signer].name}Upgradeable`,
+          path: signers[signer].path
+            .replace('.sol', 'Upgradeable.sol')
+            .replace('/contracts/', '/contracts-upgradeable/'),
+        });
```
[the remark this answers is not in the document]
ernestognw: These versions exist: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/tree/master/contracts/utils/cryptography/signers

ericglau (commented):
When upgradeability is enabled, `account.ts` manually determines which contracts should come from the upgradeable vs. vanilla library.  However, a number of vanilla contracts are not using the upgradeable variants even when they are available.

For example, in `account.ts` line 246, it has:
```
    {
      name: 'EIP712',
      path: '@openzeppelin/contracts/utils/cryptography/EIP712.sol',
    },
```
and does not use the upgradeable version [EIP712Upgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v5.4.0/contracts/utils/cryptography/EIP712Upgradeable.sol).  

What is the strategy for determining whether an upgradeable import should be used?

Note: In this case, EIP712 has a constructor so would fail validations if used with Upgrades Plugins, even if constructors are allowed on the generated contract.  We could update the tag for the generated contract according to https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1184 to allow this, but if we can avoid vanilla contracts with constructors in the first place, then that would not be necessary.

packages/core/solidity/src/account.test.ts.md
```diff
@@ -372,37 +372,55 @@ Generated by [AVA](https://avajs.dev).
...
-    import {ERC7739} from "@openzeppelin/community-contracts/utils/cryptography/ERC7739.sol";␊
-    import {SignerERC7702} from "@openzeppelin/community-contracts/utils/cryptography/SignerERC7702.sol";␊
+    import {ERC7739} from "@openzeppelin/contracts/utils/cryptography/signers/draft-ERC7739.sol";␊
+    import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";␊
+    import {UUPSUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";␊
     ␊
-    contract CustomAccountWithSignerERC7702 is Account, EIP712, ERC7739, SignerERC7702 {␊
-        constructor() EIP712("CustomAccount with SignerERC7702", "1") {}␊
+    contract CustomAccount is Initializable, Account, EIP712, ERC7739, UUPSUpgradeable {␊
+        constructor() EIP712("CustomAccount", "1") {}␊
```
ericglau, line 390: Should this disable initializers and add the annotation to allow constructors?

packages/core/solidity/src/account.test.ts.md
```diff
@@ -605,185 +721,154 @@ Generated by [AVA](https://avajs.dev).
...
+    import {ERC7739} from "@openzeppelin/contracts/utils/cryptography/signers/draft-ERC7739.sol";␊
+    import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";␊
     ␊
-    /// @custom:oz-upgrades-unsafe-allow constructor␊
-    contract CustomAccountWithSignerECDSA is Initializable, Account, EIP712, ERC7739, SignerECDSA {␊
-        constructor() EIP712("CustomAccount with SignerECDSA", "1") {␊
-            _disableInitializers();␊
-        }␊
+    contract CustomAccount is Initializable, Account, EIP712, ERC7739 {␊
+        constructor() EIP712("CustomAccount", "1") {}␊
```
ericglau, line 845: Should this disable initializers and add the annotation to allow constructors?

packages/core/solidity/src/account.test.ts.md
ernestognw: Yes, great catch, thanks! I added the `_disableInitializer` and natspec tag accordingly.

packages/core/solidity/src/account.test.ts.md
ernestognw: Same as https://github.com/OpenZeppelin/contracts-wizard/pull/609#discussion_r2289594925

ernestognw:
> What is the strategy for determining whether an upgradeable import should be used?

Right, so we're only avoiding EIP712Upgradeable for the accounts on purpose. The rationale is that the domain is still safe if it's defined in the implementation and it would be cheaper to access since it's cached in immutable storage. Although the account would work correctly if it uses EIP712Upgradeable, it would require reinitializing the domain.
 
> Note: In this case, EIP712 has a constructor so would fail validations if used with Upgrades Plugins, even if constructors are allowed on the generated contract. We could update the tag for the generated contract according to [OpenZeppelin/openzeppelin-upgrades#1184](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1184) to allow this, but if we can avoid vanilla contracts with constructors in the first place, then that would not be necessary.

Yeah I think that should work since EIP712 already is tagged with `@custom:oz-upgrades-unsafe-allow state-variable-immutable`. I understand we need to fix this TODO: https://github.com/ericglau/openzeppelin-upgrades/commit/297b8a430d076d6c50269808d482569823bef9b8. Right?

packages/core/solidity/src/account.ts
```diff
@@ -170,12 +182,55 @@
...
+      );
+      c.addFunctionCode(`_setThreshold(${signerArgs[opts.signer][2]!.name});`, fn);
+      break;
+    case 'ECDSA':
+    case 'P256':
+    case 'RSA':
+      c.addFunctionCode(`_setSigner(${signerArgs[opts.signer].map(({ name }) => name).join(', ')});`, fn);
+      break;
+  }
+}
```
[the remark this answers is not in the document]
ericglau: Initializable is available in the non-upgradeable library as well.

ericglau (approved):
@ernestognw Can you please review a few changes that I made:
- Changed `UUPSUpgradeable` to come from `contracts-upgradeable`
- Changed the constructor comment for Account/signers to use `/// @custom:oz-upgrades-unsafe-allow-reachable constructor` (currently it behaves the same as without "reachable", but when https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1184 is fixed in the Upgrades Plugins, it will allow EIP712 to pass validations.)

Also, the following are also coming from vanilla contracts even though there are upgradeable variants.  It looks fine to me as I don't see any storage usage, but please confirm if this is intentional:
- ERC1155Holder
- ERC721Holder

The rest looks great to me.  Amazing work!

LFG 🚀

ericglau:
@SocketSecurity ignore npm/prettier@3.6.2
Looks like a false positive.

packages/core/solidity/src/account.test.ts.md
```diff
@@ -372,37 +372,5843 @@ Generated by [AVA](https://avajs.dev).
...
+        // The following functions are overrides required by Solidity.␊
+    ␊
+        function _validateUserOp(PackedUserOperation calldata userOp, bytes32 userOpHash)␊
+            internal␊
+            override(Account, AccountERC7579Upgradeable)␊
+            returns (uint256)␊
+        {␊
+            return super._validateUserOp(userOp, userOpHash);␊
+        }␊
+    }␊
```
Amxx, line 712: This accoutn needs an initializer that installs a first module. Otherwise, its just bricked at construction.
This is particularly visible in this case, but I believe its all upgradeable accounts that are missing an initializer. In most cases the initializer would be empty, so we don't really care, but in cases where the initializer is actually needed that is more serious.

Amxx: Note: we have the same issue with the non-upgradeable version. If we don't have a signer, and just use 7579, the constructor doesn't install any "initial module"

ericglau (approved):

ericglau:
@SocketSecurity ignore npm/undici@5.29.0

Amxx (approved):

PR #619, opened 2025-08-08
Use ERC20Bridgeable from Contracts 5.4.0

### LINKED ISSUE #592: Update ERC20Bridgeable to use OpenZeppelin Contracts v5.4.0 when released
ERC20's Cross-Chain Bridging options are using ERC20Bridgeable from OpenZeppelin Community Contracts.  Change it to use it from OpenZeppelin Contracts v5.4.0 when that is released instead.  It is currently [available in v5.4.0-rc1](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.4.0-rc.1/contracts/token/ERC20/extensions/draft-ERC20Bridgeable.sol).

ericglau:
Fixes #592 

- For ERC20 cross chain bridging, uses ERC20Bridgeable from Contracts 5.4.0 instead of Community Contracts.
  - Enables upgradeability when cross chain bridging is enabled, since ERC20Bridgeable is included in the upgradeable variants.
  - For custom bridges, uses a state variable instead of immutable, since the bridge address should be set in the initializer for the upgradeable case.
- Updates the minimum required Contracts version to 5.4.0

Note: This PR does not address other contracts that were removed from Community Contracts. Those should be addressed by #618


ernestognw:
I'm not entirely sure how to fix the MCP error. It's failing because the wizard version installed in packages/mcp doesn't have the upgradeable ERC20Bridgeable version. Does that require a releasE?

ernestognw (approved):

CoveMB (approved):

Issue #621, opened 2025-08-11
Add git commit hash in contract comments for Community Contracts


ericglau:
In the Solidity Wizard, certain options use imports from [OpenZeppelin Community Contracts](https://github.com/OpenZeppelin/openzeppelin-community-contracts/).  For example, see Stablecoin with Limitations.

The Community Contracts is unversioned, but Wizard's lockfile [pins a specific git commit hash](https://github.com/OpenZeppelin/contracts-wizard/blob/374d3253c97d8151f196dc871f661e639f9c1c13/yarn.lock#L823) of Community Contracts which has been tested with the generated code.  Newer versions of Community Contracts may not work with Wizard's output until Wizard's code generation is updated to match.

In the generated contract's comments, we should include the git commit hash of Community Contracts that has been tested with Wizard, so that users know which version is compatible.

Also consider providing information somewhere on how to install the supported version, by providing an example NPM command (see https://docs.npmjs.com/cli/v11/configuring-npm/package-json#git-urls-as-dependencies) and/or forge command (see https://getfoundry.sh/forge/reference/install/ using `@<commit>`)

PR #627, opened 2025-08-12
Add git commit hash in output comments for Community Contracts

ericglau:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/621

- Adds a comment for the compatible git commit hash for Community Contracts imports
- Updates the links for Community Contracts imports to the specific git commit's version of each imported Solidity file.

Note: This does not include hardhat/forge instructions on how to install a specific commit.  It only adds a comment to the contract code, which is not the right context for additional instructions.


packages/core/solidity/src/print.ts
```diff
@@ -54,6 +56,15 @@ export function printContract(contract: Contract, opts?: Options): string {
...
 }
 
+function printCompatibleLibraryVersions(contract: Contract): string[] {
+  const lines: string[] = [];
+  lines.push(`// Compatible with OpenZeppelin Contracts ${compatibleContractsSemver}`);
+  if (importsCommunityContracts(contract)) {
+    lines.push(`// Compatible with OpenZeppelin Community Contracts commit ${getCommunityContractsGitCommit()}`);
+  }
+  return lines;
+}
```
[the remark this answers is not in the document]
ericglau: done

packages/core/solidity/src/print.ts
```diff
@@ -54,6 +56,20 @@ export function printContract(contract: Contract, opts?: Options): string {
   );
 }
 
+function printCompatibleLibraryVersions(contract: Contract): string[] {
```
ernestognw, line 59: Are we testing this function somehow? Not sure if it's worth it, though.

packages/core/solidity/src/print.ts
```diff
@@ -54,6 +56,20 @@ export function printContract(contract: Contract, opts?: Options): string {
...
+  lines.push(`// Compatible with OpenZeppelin Contracts ${compatibleContractsSemver}`);
+  if (importsCommunityContracts(contract)) {
+    try {
+      const commit = getCommunityContractsGitCommit();
+      lines.push(`// Compatible with OpenZeppelin Community Contracts commit ${commit}`);
+    } catch (e) {
+      console.error(e);
+    }
+  }
+  return lines;
```
ernestognw, line 70: Let's try to use only 1 row, reading the same twice feels off

```suggestion
  const lines: string[] = [];
  let compat = `// Compatible with OpenZeppelin Contracts ${compatibleContractsSemver}`;
  if (importsCommunityContracts(contract)) {
    try {
      const commit = getCommunityContractsGitCommit();
      compat += `and Community Contracts commit ${commit}`
      lines.push(`// Compatible with OpenZeppelin Community Contracts commit ${commit}`);
    } catch (e) {
      console.error(e);
    }
  }
  lines.push(compat);
  return lines;
```

packages/core/solidity/src/print.ts
```diff
@@ -54,6 +56,20 @@ export function printContract(contract: Contract, opts?: Options): string {
   );
 }
 
+function printCompatibleLibraryVersions(contract: Contract): string[] {
+  const lines: string[] = [];
+  lines.push(`// Compatible with OpenZeppelin Contracts ${compatibleContractsSemver}`);
+  if (importsCommunityContracts(contract)) {
+    try {
+      const commit = getCommunityContractsGitCommit();
+      lines.push(`// Compatible with OpenZeppelin Community Contracts commit ${commit}`);
```
ernestognw, line 65: I think it's worth slicing the commit to only 7 characters. Should be enough for a valid URL


```suggestion
      lines.push(`// Compatible with OpenZeppelin Community Contracts commit ${commit.slice(0, 7)}`);
```

packages/ui/src/solidity/inject-hyperlinks.ts
```diff
@@ -2,17 +2,32 @@ import { version as contractsVersion } from '@openzeppelin/contracts/package.jso
...
-  return code
-    .replace(
-      contractsRegex,
-      `&quot;<a class="import-link" href="https://github.com/OpenZeppelin/openzeppelin-$2blob/v${contractsVersion}/contracts/$3" target="_blank" rel="noopener noreferrer">$1$2$3</a>&quot;`,
-    )
-    .replace(
-      communityContractsRegex,
-      `&quot;<a class="import-link" href="https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/$3" target="_blank" rel="noopener noreferrer">$1$2$3</a>&quot;`,
-    );
+  const compatibleCommunityContractsRegexSingle = /OpenZeppelin Community Contracts commit ([a-f0-9]{40})/;
```
ernestognw, line 10: Accordingly, if we use a single line

```suggestion
  const compatibleCommunityContractsRegexSingle = /Community Contracts commit ([a-f0-9]{40})/;
```

packages/core/solidity/src/print.ts
ericglau: Truncating could lead to ambiguity in the unlikely event that multiple commits start with the same chars.

Instead, changed as follows:
- Define only 7 chars in the package.json devDependency -- if there is ambiguity at this point, yarn install gives an error.  In that case, we can use more characters (up to 40).
- Update regexes to allow the comment to contain between 7-40 chars.  GitHub URLs support that entire range.

packages/core/solidity/src/utils/community-contracts-git-commit.ts
```diff
@@ -0,0 +1,25 @@
...
+    );
+  }
+  const hash = split[1]!;
+  if (!/^[0-9a-f]{40}$/.test(hash)) {
+    throw new Error(
+      `Expected git commit hash for package dependency ${dependencyName} to have 40 hex chars, but got ${hash}`,
+    );
+  }
+  return hash;
+}
```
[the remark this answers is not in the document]
ericglau: The current way is sufficient since we control the format used in package.json.  This can be updated later if needed.

packages/core/solidity/src/print.ts
ernestognw: Yeah that makes sense. I agree it's better to have a check, and we can reuse the package.json with yarn install for that.

Thanks!

packages/core/solidity/src/utils/imports-libraries.ts
```diff
@@ -0,0 +1,5 @@
+import type { Contract } from '../contract';
+
+export function importsCommunityContracts(contract: Contract) {
```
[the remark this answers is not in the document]
ericglau: The s in imports implies boolean

PR #632, opened 2025-08-14
Fix MCP banner horizontal viewport overflow

gonzaotc:
Currently, the banner is slightly and laterally overflowing the user viewport (screen size), enabling lateral scroll in the wizard website, which uncenters the layout and hides the OpenZeppelin Logo, providing a less stable layout experience.

Fixed with this small padding change. 

Screenshot taken at scrolling horizontally prior to fix: 
image


ericglau (approved):
Nice fix, thanks!

PR #638, opened 2025-08-19
Cairo: Support v3.0.0-alpha.1

ericnordelo:
Updates the interface dependencies.

PR #640, opened 2025-08-19
Cairo: Use await async in update-scarb-project

ericglau:
The async function `updateScarbProject` did not use await when calling other functions.  This causes the script to return while the other functions are still waiting, which leads to race conditions for anything that relies on the script such as the compile cairo project workflows.

Originally reported in https://github.com/OpenZeppelin/contracts-wizard/pull/638#pullrequestreview-3131618650


CoveMB (approved):

CoveMB (approved):
Maybe we could add a test that would have catch this?

PR #658, opened 2025-09-08
Unify upgradeability of Account with the other types of contract

Amxx:
This PR addresses an inconsistency introduced in #609
This is done by reverting a significant portion the changes to upgradeability transformation introduced in #609

## About the produced solidity code

Non-upgradeable accounts, that include a signer, are currently quite strange:
- They are Initializable, which should not be the case of non-upgradeable contracts
- They have a constructor that initializes the signer (ok) and do a `_disableInitializer()`
- They have a signer initializer, which can never be called on the contract itself (because of the disable at construction)

In fact they are in a mixed state between beeing non upgradeable, and being an implementation for proxies.

With this PR, non-upgradeable accounts are
- not Initializable
- don't have a `_disableInitializer` call
- don't have a public initializer.

## How its implemented

- The changes to `contract.ts` to its state before #609. In particular `shouldAutoTranspileImports`, `shouldInstallContractsUpgradeable` and `shouldUseUpgradesPluginsForProxyDeployment` are removed in favor of the old `upgradealbe`
- I reverted most changes to `setUpgradeable.ts`, and use a simpler version of `setUpgradeableAccount` (similar to the existing variants for generic and governor)
- I use the pre-existing `ReferencedContract.transpiled: boolean | undefined` to explicit which parents should and shouldn't be transpiled.
- I change the account and signer logic to not distinguish upgradeable/non-upgradeable case, and let the wizard take care of that (turning constructors in initializers)

## Summary by Sourcery

Revert the experimental upgradeability transformation from #609 and restore the original upgradeable/non-upgradeable account behavior using a single `upgradeable` flag.

Enhancements:
- Restore the single `upgradeable` boolean in contract definitions and remove the specialized `shouldAutoTranspileImports`, `shouldInstallContractsUpgradeable`, and `shouldUseUpgradesPluginsForProxyDeployment` flags
- Revert the majority of changes in `set-upgradeable.ts` and `contract.ts` to their pre-#609 state and simplify the upgradeable account pattern
- Update constructor/initializer generation so non-upgradeable accounts no longer implement `Initializable`, call `_disableInitializers()`, or expose public initializers
- Unify account and signer logic to treat upgradeable and non-upgradeable cases the same, delegating initializer emission to the wizard

### DIFF
--- a/packages/core/solidity/src/zip-foundry.ts
+++ b/packages/core/solidity/src/zip-foundry.ts
@@ -6,6 +6,7 @@ import SOLIDITY_VERSION from './solidity-version.json';
 import contracts from '../openzeppelin-contracts';
 import type { Lines } from './utils/format-lines';
 import { formatLinesWithSpaces, spaceBetween } from './utils/format-lines';
+import type { Upgradeable } from './set-upgradeable';
 
 function getHeader(c: Contract) {
   return [`// SPDX-License-Identifier: ${c.license}`, `pragma solidity ^${SOLIDITY_VERSION};`];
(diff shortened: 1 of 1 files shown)


packages/core/solidity/src/zip-hardhat.ts
```diff
@@ -119,9 +119,14 @@ function getAddressArgs(c: Contract): string[] {
 }
 
 function getDeploymentCall(c: Contract, args: string[]): string {
-  return c.shouldUseUpgradesPluginsForProxyDeployment
+  // TODO: remove that selector when the upgrades plugin supports @custom:oz-upgrades-unsafe-allow-reachable
+  const useUpgradePlugin = c.parents.find(p => ['EIP712'].includes(p.contract.name)) == undefined;
```
ericglau, line 123: This can be named and inverted to something like `useAllowConstructorFlag`. Since the plugin is being used in both cases.

packages/core/solidity/src/zip-foundry.ts
```diff
@@ -16,8 +16,17 @@ const test = (c: Contract, opts?: GenericOptions) => {
...
-      result.push('import {Upgrades} from "openzeppelin-foundry-upgrades/Upgrades.sol";');
+    if (c.upgradeable) {
+      // TODO: remove that selector when the upgrades plugin supports @custom:oz-upgrades-unsafe-allow-reachable
+      const useUpgradePlugin = c.parents.find(p => ['EIP712'].includes(p.contract.name)) == undefined;
+
+      result.push(
+        useUpgradePlugin
+          ? 'import {Upgrades} from "openzeppelin-foundry-upgrades/Upgrades.sol";'
+          : opts?.upgradeable == 'transparent'
+            ? 'import {TransparentUpgradeableProxy} from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";'
```
ericglau, line 27: We can set `unsafeAllow = "constructor"` through the Options struct, and still use the plugin: https://docs.openzeppelin.com/upgrades-plugins/api-foundry-upgrades#Options-Options

packages/core/solidity/src/account.ts
```diff
@@ -133,6 +128,7 @@ function addERC721Holder(c: ContractBuilder, opts: AccountOptions): void {
   c.addParent({
     name: 'ERC721Holder',
     path: '@openzeppelin/contracts/token/ERC721/utils/ERC721Holder.sol',
+    transpiled: false,
```
ericglau, line 131: There is an [upgradeable variant](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v5.4.0/contracts/token/ERC721/utils/ERC721HolderUpgradeable.sol) of this.  Any reason not to use it?

packages/core/solidity/src/account.ts
```diff
@@ -141,6 +137,7 @@ function addERC1155Holder(c: ContractBuilder, opts: AccountOptions): void {
   c.addParent({
     name: 'ERC1155Holder',
     path: '@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol',
+    transpiled: false,
```
ericglau, line 140: There is an [upgradeable variant](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v5.4.0/contracts/token/ERC1155/utils/ERC1155HolderUpgradeable.sol) of this.  Any reason not to use it?

packages/core/solidity/src/account.ts
Amxx: I would argue there should not be an upgradeable version. This is stateless, and I think we should at some point remove the upgradeable version.

That being said, we can remove the `transpiled: false` flag, and use the upgradeable version. Result logic will be the same. One of the difference though, is that you'll have a `__ERC721Holder_init()` in your constructor that you don't need. So I think its clearner to use the non-upgradeable version, but I'd be ok with using the upgradeable version.

All the above apply to ERC1155Holder.

packages/core/solidity/src/zip-foundry.ts
Amxx: addressed in https://github.com/OpenZeppelin/contracts-wizard/pull/658/commits/b3e3dba32a26efa97aa6715a98e0d1d87197d340

packages/core/solidity/src/zip-hardhat.ts
Amxx: addressed in e303818

ericglau (approved):
LGTM, thanks!

I pushed some changes to clarify the parents when printing constructors/initializers, and refactored zip-foundry.ts to reduce code duplication.

ernestognw:
> I think 576 is lower priority. I'd first make sure that the Accounts are generated "properly" before adding something extra

It is a priority for the Open Accounts Framework as all accounts would require an initializer. Again, please name at least one popular use case for accounts that have an initializer instead. In fact, I consider this current implementation more strange and way more misleading than the original one.

I have a strong opinion so won't approve. For me, the correct implementation is still #653

PR #671, opened 2025-09-24
docs(core): fix grammar and parallelism

isagi-y22:


ericglau:
Thanks for the PR but I don't find this necessary.  The previous grammar with "but" is correct because the example shows how to change one of the fields from the defaults.

PR #681, opened 2025-09-29
[Solidity] Upgrade @openzeppelin/contracts to 5.5.0

### LINKED ISSUE #660: core/solidity: set UUPSUpgradeable as "transpile: false" in set-upgradeable.
In `packages/core/solidity/src/set-upgradeable.ts` make `UUPSUpgradeable` parent `transpile: false`.

- Simplify the produced bytecode, removing the call to `__UUPSUpgradeable_init();` that does nothing
- Change the dependency from UUPSUpgradeable from `@openzeppelin/contracts-upgradeable` to `@openzeppelin/contracts`, allowing us to stop transpilling it in the future (its stateless)

Amxx:
Fixes #660

### DIFF
--- a/packages/core/solidity/src/print.ts
+++ b/packages/core/solidity/src/print.ts
@@ -102,7 +102,10 @@ function printConstructor(contract: Contract, helpers: Helpers): Lines[] {
     if (helpers.upgradeable) {
       const upgradeableParents = parentsWithInitializers.filter(p => inferTranspiled(p.contract));
       // Omit Initializable and UUPSUpgradeable since they don't have explicit constructors
-      const nonUpgradeableParentsWithConstructors = contract.parents.filter(p => !inferTranspiled(p.contract) && p.contract.name !== 'Initializable' && p.contract.name !== 'UUPSUpgradeable');
+      const nonUpgradeableParentsWithConstructors = contract.parents.filter(
+        p =>
+          !inferTranspiled(p.contract) && p.contract.name !== 'Initializable' && p.contract.name !== 'UUPSUpgradeable',
+      );
       const constructor = printFunction2(
         [
           nonUpgradeableParentsWithConstructors.length > 0


packages/core/solidity/src/environments/hardhat/package.json
```diff
@@ -9,8 +9,8 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@openzeppelin/contracts": "^5.4.0",
     "@nomicfoundation/hardhat-toolbox": "^6.1.0",
-    "hardhat": "^2.16.1"
+    "@openzeppelin/contracts": "^5.5.0-rc.0",
+    "hardhat": "^2.22.0"
```
Amxx, line 14: 5.5.0-rc.0 requires Cancun support for mcopy. [This is supported by hardhat since 2.22.0](https://github.com/NomicFoundation/hardhat/blob/v2/packages/hardhat-core/CHANGELOG.md#2220).

Amxx:
@ericglau, 5.5.0 is now released !

.changeset/lazy-seals-cover.md
```diff
@@ -0,0 +1,10 @@
+---
+'@openzeppelin/wizard': patch
+'@openzeppelin/wizard-common': patch
+'@openzeppelin/contracts-mcp': patch
```
ericglau: We could bump the minor versions (major version is still 0) to more strongly signify the breaking changes. (We recently also did so for [namespaced storage](https://github.com/OpenZeppelin/contracts-wizard/releases/tag/%40openzeppelin%2Fwizard%400.9.0)).
```suggestion
'@openzeppelin/wizard': minor
'@openzeppelin/wizard-common': patch
'@openzeppelin/contracts-mcp': minor
```

packages/core/solidity/src/account.ts
```diff
@@ -286,6 +286,7 @@ const functions = {
       args: [
         { name: 'userOp', type: 'PackedUserOperation calldata' },
         { name: 'userOpHash', type: 'bytes32' },
+        { name: 'signature', type: 'bytes calldata' },
```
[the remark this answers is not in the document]
ericglau: I added a related comment to the changeset, I think it is worth specifying a breaking change for the Initializable/UUPSUpgradeable portions already.

We should also mention this part here in the changeset.

ericglau:
@SocketSecurity ignore-all
Looks like false positives.

PR #698, opened 2025-10-21
[Cairo] Support AccessControlDefaultAdminRules

immrsd:
Adds support for `AccessControlDefaultAdminRules` component by introducing an additional option to the access module: `roles-dar`. Refactors `access` to be an object with type and additional options instead of being value-based. Introduces UI controls for configuring the `AccessControlDefaultAdminRules` settings.

### DIFF
--- a/packages/core/cairo_alpha/src/scripts/update-scarb-project.ts
+++ b/packages/core/cairo_alpha/src/scripts/update-scarb-project.ts
@@ -78,6 +78,25 @@ async function updateScarbToml() {
   await fs.writeFile(scarbTomlPath, updatedContent, 'utf8');
 }
 
+function parseCliArgs(args: string[]): Record<string, string> {
+  const result: Record<string, string> = {};
+  for (let i = 0; i < args.length; i++) {
+    const arg = args[i];
+    if (!arg) continue;
+    if (arg.startsWith('--') && arg.slice(2).includes('=')) {
+      const [key, value] = arg.slice(2).split('=', 2);
+      if (key && value !== undefined) {
+        result[key] = value;
+      } else {
+        throw new Error(`Invalid argument format: ${arg}`);
+      }
+    } else {
+      throw new Error(`Invalid argument format: ${arg}. Expected --key=value`);
+    }
+  }
+  return result;
+}
+
 function parseKindSubset(value: string | undefined): KindSubset {
   if (value === undefined) {
     throw new Error(`Failed to resolve contract kind subset from 'undefined'.`);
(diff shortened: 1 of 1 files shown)


packages/core/cairo_alpha/src/common-options.ts
```diff
@@ -11,7 +12,7 @@ export const defaults: Required<CommonOptions> = {
 
 export const contractDefaults: Required<CommonContractOptions> = {
   ...defaults,
-  access: false,
+  access: { ...AccessControl.None },
```
ericnordelo: Why does this need an object? Why not just AccessControl.None?

packages/ui/src/cairo_alpha/App.svelte
```diff
@@ -100,10 +100,9 @@
 
   $: code = printContract(contract);
   $: highlightedCode = injectHyperlinks(hljs.highlight(code, { language: 'cairo' }).value);
-
   $: hasErrors = errors[tab] !== undefined;
 
-  const language = 'cairo';
+  const language = 'cairo-alpha';
```
ericnordelo: Should the language include alpha?

ericnordelo (commented):
Looking good @immrsd. Left some small comments

packages/core/cairo_alpha/src/common-options.ts
immrsd: I used this pattern to make a copy of the object to safeguard against its possible mutations. However, I've pushed changes that make `AccessControl.None` a function returning a brand-new object each time rather than a standalone object. That prevents this very issue in a better way so that you don't have to worry about object references at the call site

packages/ui/src/cairo_alpha/AccessControlSection.svelte
```diff
@@ -1,55 +1,84 @@
...
       }
     }
 
     wasRequired = required;
-    if (access !== false) {
-      defaultValueWhenEnabled = access;
+    if (accessType !== false) {
+      defaultTypeWhenEnabled = accessType;
     }
   }
```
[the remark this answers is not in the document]
immrsd: This logic works correctly and the PR brings no logical changes, just some renaming

packages/core/cairo_alpha/src/scripts/update-scarb-project.ts
```diff
@@ -107,6 +106,69 @@ function parseKindSubset(value: string | undefined): KindSubset {
...
+function parseAccessSubset(value: string | undefined): AccessSubset {
+  if (value === undefined) {
+    throw new Error(`Failed to resolve access subset from 'undefined'.`);
+  }
+  switch (value.toLowerCase()) {
+    case 'all':
+      return 'all';
+    case 'disabled':
+    case 'none':
+    case 'no':
```
ericglau: Is there a need to support these different aliases?

packages/core/cairo_alpha/src/scripts/update-scarb-project.ts
```diff
@@ -107,6 +106,69 @@ function parseKindSubset(value: string | undefined): KindSubset {
...
+      return 'disabled';
+    case 'ownable':
+      return 'ownable';
+    case 'roles':
+      return 'roles';
+    case 'roles-dar-default':
+    case 'roles_dar_default':
```
ericglau: Is there a need to support these different aliases?

packages/core/cairo_alpha/src/scripts/update-scarb-project.ts
```diff
@@ -107,6 +106,69 @@ function parseKindSubset(value: string | undefined): KindSubset {
...
+      return 'roles-dar-default';
+    case 'roles-dar-custom':
+    case 'roles_dar_custom':
```
ericglau: Is there a need to support these different aliases?

immrsd: Only to prevent mistakes with param name when running the script locally

packages/core/cairo_alpha/src/scripts/update-scarb-project.ts
```diff
@@ -107,6 +106,69 @@ function parseKindSubset(value: string | undefined): KindSubset {
...
+        const parts = argName.split('=', 2);
+        const key = parts[0];
+        const value = parts[1];
+        if (key && value !== undefined) {
+          result[key] = value;
+        } else {
+          throw new Error(`Invalid argument format: ${arg}`);
+        }
+      } else if (i + 1 < args.length) {
+        // Handle --arg value format
```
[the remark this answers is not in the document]
ericglau: That sounds fine, although I have a slight preference for being stricter (and failing early if a wrong input is provided), just so there's less code to maintain.  Same for the aliases below.  

If we do want to support different input formats or aliases, perhaps those can be refactored into helpers to make this part of the code cleaner.

PR #703, opened 2025-10-22
[Cairo] with_components macro

immrsd:
## Key Changes
- Adds support for `with_component` macro in Cairo-alpha
- Introduces `MacrosSubset` setting for contract generators
- Improves the `generate_test_project` script in Cairo-alpha


packages/ui/src/cairo_alpha/MacrosSection.svelte
```diff
@@ -0,0 +1,20 @@
...
+
+<section class="controls-section">
+  <h1>Macros</h1>
+
+  <div class="checkbox-group">
+    <label class:checked={macros.withComponents}>
+      <input type="checkbox" bind:checked={macros.withComponents} />
+      with_components
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts-cairo/alpha/macros/with_components">
+        Use the "with_components" macro for simplified contract structure.
```
ericglau: This can use html if desired:
```suggestion
        Use the <code>with_components</code> macro for simplified contract structure.
```

packages/core/cairo_alpha/src/print.ts
```diff
@@ -201,35 +218,28 @@ function printImpl(impl: Impl): Lines[] {
 }
 
 function printStorage(contract: Contract): (string | string[])[] {
-  const lines = [];
-  // storage is required regardless of whether there are components
-  lines.push('#[storage]');
-  lines.push('struct Storage {');
+  if (contract.macros.withComponents || contract.components.length === 0) {
+    // storage is required regardless of whether there are components
```
ericglau: does this comment need to be updated/removed?

packages/core/cairo_alpha/src/scripts/update-scarb-project.ts
```diff
@@ -5,44 +5,82 @@ import type { KindSubset } from '../generate/sources';
...
+        const parts = argName.split('=', 2);
+        const key = parts[0];
+        const value = parts[1];
+        if (key && value !== undefined) {
+          result[key] = value;
+        } else {
+          throw new Error(`Invalid argument format: ${arg}`);
+        }
+      } else if (i + 1 < args.length) {
+        // Handle --arg value format
```
ericglau: Do we ever pass in this format?

packages/core/cairo_alpha/src/print.ts
```diff
@@ -201,35 +218,28 @@ function printImpl(impl: Impl): Lines[] {
...
+  if (contract.macros.withComponents) {
+    return [];
   }
-  return lines;
+  const eventLines = [];
+  for (const component of contract.components) {
+    eventLines.push('#[flat]');
+    eventLines.push(`${component.event.name}: ${component.event.type},`);
+  }
+  return ['#[event]', '#[derive(Drop, starknet::Event)]', 'enum Event {', eventLines, '}'];
```
ericnordelo: We still should not print an empty event struct because it is not required.

ericnordelo (commented):
Great job @immrsd. Looking very good. Left some comments.

packages/core/cairo_alpha/src/set-access-control.ts
```diff
@@ -25,19 +25,23 @@ export function setAccessControl(c: ContractBuilder, access: Access): void {
...
+            embed: true,
             value: 'AccessControlComponent::AccessControlWithDelayImpl<ContractState>',
           });
         } else {
           c.addImplToComponent(components.AccessControlComponent, {
             name: 'AccessControlMixinImpl',
+            embed: true,
             value: 'AccessControlComponent::AccessControlMixinImpl<ContractState>',
           });
           c.addInterfaceFlag('ISRC5');
```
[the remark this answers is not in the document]
immrsd: Not needed

ericglau (approved):
LGTM.  However, having two separate copies of the test suites (with components on and off) seems excessive in terms of maintenance overhead.

Would it be sufficient to just keep all existing tests with components OFF by default, but add two test cases for each contract with components ON?  For example, with no features enabled and components ON, and with all features enabled and components ON.

I think the above could either be part of this or a separate PR.  (Feel free to merge if planned for a separate PR)

Issue #383, opened 2024-09-03
Use namespaced storage for state variables when upgradeable


ericglau:
From https://wizard.openzeppelin.com/#erc721,  enable "Mintable -> Auto Increment Ids" and "Upgradeability".  This adds state variable `uint256 private _nextTokenId;`to the code, and this can be enhanced to use namespaced storage as a best practice for upgradeable contracts.

(This is currently the only state variable added by Wizard, but this should apply for any state variable(s) being added by Wizard when upgradeability is enabled.)

ericglau:
We should consider what to use as the namespace id here.  Since this namespace belongs in the user's contract rather than in a contract library, this id should be somewhat unique for each user and reflect something that they can add variables to.  In that case, we might want to allow users to customize this with a prefix like `myproject`, which results in a namespace id like `myproject.MyToken`.

However, namespace id conflicts could occur if different users inherit each other's code that uses the same namespace id if they don't change the defaults.  The Upgrades Plugins will detect this though, and I'm not sure how likely this conflict scenario would be in the first place.

Ideally, users would just see a namespace in their contract and then use https://github.com/OpenZeppelin/openzeppelin-vscode to edit the namespace id to whatever they want :)  So perhaps having a generic default is enough.

typicalHuman:
I believe this issue can be closed.

PR #704, opened 2025-10-24
Use namespaced storage for variables when upgradeable

typicalHuman:
Use namespaced storage for variables when upgradeability is enabled:
  - For ERC-20, use namespaced storage for `tokenBridge` when cross-chain bridging is set to `'custom'` and upgradeability is enabled.
  - For ERC-721, use namespaced storage for `_nextTokenId` when mintable, auto increment IDs, and upgradeability are enabled.

https://github.com/OpenZeppelin/contracts-wizard/issues/383


packages/core/solidity/src/erc721.ts
```diff
@@ -174,14 +175,28 @@ function addBurnable(c: ContractBuilder) {
...
+      c.addStructVariable(storageStruct, 'uint256 _nextTokenId;');
+      c.addVariable(
+        `bytes32 private constant ${namespacedStorageName} = ${generateNamespacesStorageSlot(getNamespacedStorageName(c.name))};`,
+      );
+      c.addFunctionCode(`assembly { $.slot :=  ${namespacedStorageName}}`, storageFn);
+
+      c.addFunctionCode('Storage storage $ = _getStorage();', fn);
+      c.addFunctionCode('uint256 tokenId = $._nextTokenId++;', fn);
+      c.addFunctionCode('_safeMint(to, tokenId);', fn);
+    }
```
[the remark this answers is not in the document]
typicalHuman: Implementation take from - https://erc7201.cc/example.main

packages/core/solidity/package.json
```diff
@@ -30,10 +30,11 @@
     "ava": "^6.0.0",
     "hardhat": "^2.22.11",
     "jszip": "^3.6.0",
+    "keccak": "^3.0.4",
```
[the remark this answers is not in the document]
typicalHuman: Almost all dependencies are stored in devDependencies, so I believe it's fine

packages/core/solidity/src/erc721.ts
```diff
@@ -174,14 +175,28 @@ function addBurnable(c: ContractBuilder) {
   });
 }
 
-function addMintable(c: ContractBuilder, access: Access, incremental = false, uriStorage = false) {
+function addMintable(c: ContractBuilder, access: Access, incremental = false, uriStorage = false, uups = false) {
```
ericglau: Namespaced storage should also apply if `upgradeable` is `transparent`, e.g. if `allOpts.upgradeable` is truthy.

packages/core/solidity/src/erc721.ts
```diff
@@ -264,3 +279,24 @@ function getMintFunction(incremental: boolean, uriStorage: boolean): BaseFunctio
 
   return fn;
 }
+
+function getStorageFunction(): BaseFunction {
+  const fn: BaseFunction = {
+    name: '_getStorage',
+    kind: 'internal' as const,
```
ericglau:
```suggestion
    kind: 'private' as const,
```

packages/core/solidity/src/erc721.ts
```diff
@@ -264,3 +279,24 @@ function getMintFunction(incremental: boolean, uriStorage: boolean): BaseFunctio
...
+    args: [],
+    returns: ['Storage storage $'],
+  };
+
+  return fn;
+}
+
+function getStorageStruct(name: string) {
+  const struct: ContractStruct = {
+    name: 'Storage',
```
ericglau: Similar to the above, prefix the struct type with the contract name e.g. `${name}Storage`

packages/core/solidity/src/erc721.ts
```diff
@@ -174,14 +175,28 @@ function addBurnable(c: ContractBuilder) {
...
+      c.addFunctionCode('_safeMint(to, tokenId);', fn);
+    } else {
+      const storageFn = getStorageFunction();
+      const storageStruct = getStorageStruct(c.name);
+      const namespacedStorageName = `${c.name.toUpperCase()}_STORAGE_LOCATION`;
+      c.addStructVariable(storageStruct, 'uint256 _nextTokenId;');
+      c.addVariable(
+        `bytes32 private constant ${namespacedStorageName} = ${computeNamespacedStorageSlot(getNamespaceId(c.name))};`,
+      );
+      c.addFunctionCode(`assembly { $.slot :=  ${namespacedStorageName}}`, storageFn);
```
ericglau: The formatting/spacing is inconsistent between the opening and closing assembly brackets, and within the := assignment.  Can you either make it consistently single-spaced, and/or use multiple lines (e.g. multiple `c.addFunctionCode` calls)

packages/core/solidity/src/erc721.ts
```diff
@@ -264,3 +279,24 @@ function getMintFunction(incremental: boolean, uriStorage: boolean): BaseFunctio
 
   return fn;
 }
+
+function getStorageFunction(): BaseFunction {
+  const fn: BaseFunction = {
+    name: '_getStorage',
```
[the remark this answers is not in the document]
typicalHuman: Done

packages/core/solidity/src/erc721.ts
```diff
@@ -264,3 +279,24 @@ function getMintFunction(incremental: boolean, uriStorage: boolean): BaseFunctio
...
+    kind: 'internal' as const,
+    mutability: 'pure',
+    args: [],
+    returns: ['Storage storage $'],
```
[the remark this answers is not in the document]
typicalHuman: Done

packages/core/solidity/src/erc721.ts
typicalHuman: Done

packages/core/solidity/src/erc721.ts
```diff
@@ -174,14 +175,28 @@ function addBurnable(c: ContractBuilder) {
...
+    if (!uups) {
+      c.addVariable('uint256 private _nextTokenId;');
+      c.addFunctionCode('uint256 tokenId = _nextTokenId++;', fn);
+      c.addFunctionCode('_safeMint(to, tokenId);', fn);
+    } else {
+      const storageFn = getStorageFunction();
+      const storageStruct = getStorageStruct(c.name);
+      const namespacedStorageName = `${c.name.toUpperCase()}_STORAGE_LOCATION`;
+      c.addStructVariable(storageStruct, 'uint256 _nextTokenId;');
+      c.addVariable(
```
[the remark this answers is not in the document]
typicalHuman: Done

ericglau (approved):
We'll do another review internally before proceeding, but this LGTM, thanks!

packages/core/solidity/package.json
```diff
@@ -21,6 +21,9 @@
     "test:watch": "ava --watch",
     "update-env": "rm ./src/environments/hardhat/package-lock.json && npm install --package-lock-only --prefix ./src/environments/hardhat && rm ./src/environments/hardhat/upgradeable/package-lock.json && npm install --package-lock-only --prefix ./src/environments/hardhat/upgradeable"
   },
+  "dependencies": {
+    "ethereum-cryptography": "^3.2.0"
+  },
```
ernestognw: Few questions:

- why is this needed in `dependencies` rather than `devDependencies`?
- any particular reason to prefer this package? Since we have `@openzeppelin/contracts` in devDependencies, then both `@ethersproject/keccak256` and `ethereum-cryptography` should be available

typicalHuman: Yes, you're right, I've updated implementation so it would use `@ethersproject` dependencies instead.

ernestognw: > This is in dependencies because it is needed at runtime to calculate erc7201 slot locations based on user inputted parameters.

Ah that makes sense, so it needs to be bundled along with the `core` package for distribution.

>  I just chose ethereum-cryptography because it is maintained under the Ethereum GitHub org and was audited.

I took a look and it hasn't had changes since 6 months ago, but I also see that `@ethersproject/keccak256` hasn't been updated in 8 months. On a second thought I think it makes sense to keep `ethereum-cryptography` then. Sorry for the inconvenience @typicalHuman 😅

ericglau: I've pushed reverts for the last 2 commits to restore it to the above with `ethereum-cryptography` in `dependencies`, and verified that it is required in dependencies.  Sorry for the confusion.

ernestognw (approved):

PR #718, opened 2025-11-03
Add WebAuthn signer to Accounts

ernestognw:
After Solidity Contracts 5.5, the WebAuthn Signer is available and we were missing the option for it.


packages/core/solidity/src/account.test.ts
```diff
@@ -61,7 +61,7 @@ function format(upgradeable: false | 'uups' | 'transparent') {
   }
 }
 
-for (const signer of [false, 'ECDSA', 'EIP7702', 'P256', 'RSA', 'Multisig', 'MultisigWeighted'] as const) {
+for (const signer of [false, 'ECDSA', 'EIP7702', 'P256', 'WebAuthn', 'RSA', 'Multisig', 'MultisigWeighted'] as const) {
```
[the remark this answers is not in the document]
ericglau: Should this just iterate over the `SignerOptions` const from `packages/core/solidity/src/signer.ts`?

packages/ui/src/solidity/AccountControls.svelte
```diff
@@ -216,6 +216,13 @@
         hardware security modules that use RSA keys.
       </HelpTooltip>
     </label>
+    <label class:checked={opts.signer === 'WebAuthn'}>
+      <input type="radio" bind:group={opts.signer} value="WebAuthn" />
+      WebAuthn
+      <HelpTooltip link="https://docs.openzeppelin.com/contracts/5.x/api/utils/cryptography#WebAuthn">
```
ericglau: This WebAuthn section doesn't appear to be present on the linked page.

packages/core/solidity/src/signer.ts
```diff
@@ -34,6 +43,22 @@ export function addSigner(c: ContractBuilder, signer: SignerOptions, upgradeable
...
+      c.addImportOnly({
+        name: 'AbstractSigner',
+        path: '@openzeppelin/contracts/utils/cryptography/signers/AbstractSigner.sol',
+        transpiled: false,
+      });
+      c.addOverride({ name: 'AbstractSigner', transpiled: false }, signerFunctions._rawSignatureValidation);
+      c.addOverride({ name: 'SignerP256' }, signerFunctions._rawSignatureValidation);
+      break;
+    }
   }
```
ericglau: It would help to add a static type assertion here in case a new signer is added in the future and we forget to add a `case` above.
Something like:
```
    default: {
      const _: never = signer;
      throw new Error('Unknown signer');
    }
```

packages/ui/api/ai-assistant/function-definitions/solidity.ts
```diff
@@ -211,7 +211,7 @@ export const solidityAccountAIFunctionDefinition = {
       signer: {
         anyOf: [
           { type: 'boolean', enum: [false] },
-          { type: 'string', enum: ['ECDSA', 'EIP7702', 'P256', 'RSA', 'Multisig', 'MultisigWeighted'] },
+          { type: 'string', enum: ['ECDSA', 'EIP7702', 'P256', 'WebAuthn', 'RSA', 'Multisig', 'MultisigWeighted'] },
```
ericglau: nit: these could be ordered alphabetically

ericglau (changes requested):

packages/ui/src/solidity/AccountControls.svelte
ernestognw: Yeah that's right. I just DM'd @stevedylandev to see if we can trigger the build on docs again to publish them

ernestognw: It's now available
https://docs.openzeppelin.com/contracts/5.x/api/utils/cryptography#WebAuthn

Issue #716, opened 2025-11-03
Soldity Stablecoin/RWA - Replace ERC20Custodian with ERC20Freezable


ericglau:
From Stablecoin or Real-World Asset tabs, Custodian option uses [ERC20Custodian](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Custodian.sol) which is deprecated.

Replace this with [ERC20Freezable](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Freezable.sol).

This may require some minor design in terms of what the resulting contract should look like and how it works with access control.

PR #719, opened 2025-11-04
Replace ERC20Custodian with ERC20Freezable

typicalHuman:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/716


packages/core/solidity/src/stablecoin.test.ts
```diff
@@ -86,8 +86,8 @@ testStablecoin('stablecoin permit', {
   permit: true,
 });
 
-testStablecoin('stablecoin custodian', {
```
ernestognw: Realizing we're missing a "managed" test:

```
testStablecoin('stablecoin mintable with roles', {
  access: 'managed',
});
```

ernestognw (approved):
Minor nit

packages/core/solidity/src/stablecoin.test.ts
typicalHuman: done

PR #723, opened 2025-11-07
[Stellar] Use constructor arguments in Stellar contracts

ElliotFriend:
Stellar supports `__constructor` functions in a smart contract, but most of the wizard contracts still hard-code the token name and symbol in the constructor function. this makes sharing the contract executables between multiple tokens more difficult, and will lead to more fees paid long-term.

This PR aims to put all the hard-coded bits of these wizard contracts into the constructor arguments. It also provides a helpful deploy command in the outputted Rust code, instructing the user how they can deploy the contract with their configured name, symbol, etc. for all the constructor arguments required.

I started in on adding some tests to cover the new functionality, but wanted to get this PR out there for a sanity check first. I can add more tests, if it seems like i'm moving in the right direction.

If there's any changes that need to be made, I'm definitely happy to do so. Just let me know 😁


ElliotFriend:

Issue #695, opened 2025-10-17
Add Solidity Wizard API function to get versioned remappings


ericglau:
### Overview

In the [programmatic API](https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/core/solidity/README.md), the `print` function creates a contract with imports that are not versioned.  For the reasons described in https://github.com/OpenZeppelin/contracts-wizard/issues/692#issuecomment-3412686360, these should remain unversioned, but callers such as Remix should add remappings to pin dependencies to specific versions.

---

### Enhancement Request

Provide function(s) in the API that when given specific contract options, returns remappings that map unversioned import prefixes to versioned import prefixes.

For example, the interface may look something like:
```
getVersionedRemappings: (opts?: Options) => string[];
```

When `opts.upgradeable` is falsy, this should return a string array with one element, e.g.:
```
[
  "@openzeppelin/contracts/=@openzeppelin/contracts@5.4.0/"
]
```

When `opts.upgradeable` is truthy, this should return a string array with two elements (for both non-upgradeable and upgradeable libraries), e.g.:
```
[
  "@openzeppelin/contracts/=@openzeppelin/contracts@5.4.0/",
  "@openzeppelin/contracts-upgradeable/=@openzeppelin/contracts-upgradeable@5.4.0/"
]
```

Callers such as Remix should append these elements as lines in `remappings.txt`.  End users can update the versions in that file if they want their code to use newer versions of OpenZeppelin Contracts, when new versions are available.

Chesblaw:
@ericglau sup can you assign me

typicalHuman:
https://github.com/OpenZeppelin/contracts-wizard/pull/724

PR #724, opened 2025-11-08
Add Solidity Wizard API function to get versioned remappings

typicalHuman:
Fixes https://github.com/OpenZeppelin/contracts-wizard/issues/695


.changeset/long-trams-tease.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openzeppelin/wizard': minor
+'@openzeppelin/wizard-common': minor
```
ericglau: There are no changes to wizard-common, and the ones for wizard (even though this is a new feature) should be patch since they are not breaking:
```suggestion
'@openzeppelin/wizard': patch
```

packages/core/solidity/src/utils/version.ts
```diff
@@ -2,3 +2,7 @@
  * Semantic version string representing of the minimum compatible version of Contracts to display in output.
  */
 export const compatibleContractsSemver = '^5.5.0';
+
+export function getRawCompatibleContractsSemver() {
```
ericglau: The caller (e.g. in get-versioned-remappings.ts) can just read `contracts.version` from `import contracts from '../openzeppelin-contracts';`.  That corresponds to the actual latest version that was used during the build, which can be higher than the minimum semantic version above.

See `packages/core/solidity/src/print-versioned.ts` for example.

In that case, the changes in this version.ts file can be reverted.

typicalHuman: Done

packages/core/solidity/src/api.ts
```diff
@@ -37,13 +37,19 @@ import {
...
+import { getVersionedRemappings } from './get-versioned-remappings';
 
 export interface WizardContractAPI<Options extends CommonOptions> {
   /**
    * Returns a string representation of a contract generated using the provided options. If opts is not provided, uses `defaults`.
    */
   print: (opts?: Options) => string;
 
+  /**
+   * Returns remappings that map unversioned import prefixes to versioned import prefixes.
```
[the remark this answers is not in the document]
typicalHuman: Done

PR #730, opened 2025-11-11
[Stellar] Add link to UI builder for stellar

CoveMB:
Add link to UI builder for stellar

### DIFF
--- a/packages/ui/public/index.html
+++ b/packages/ui/public/index.html
@@ -112,7 +112,7 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
       </div>
 
       <div class="flex items-center">
-        <a href="https://builder.openzeppelin.com?ecosystem=evm" target="_blank" rel="noopener noreferrer" class="bg-[#3d3b5d] text-white px-4 py-2 rounded-full text-sm font-medium flex items-center gap-2 hover:bg-[#4a4872] transition-colors no-underline whitespace-nowrap">
+        <a href="https://builder.openzeppelin.com" target="_blank" rel="noopener noreferrer" class="bg-[#3d3b5d] text-white px-4 py-2 rounded-full text-sm font-medium flex items-center gap-2 hover:bg-[#4a4872] transition-colors no-underline whitespace-nowrap">
           <span>Try the UI Builder</span>
           <img src="/icons/up-right-from-square.svg" alt="up-right-from-square" class="arrow-white">
         </a>


packages/ui/public/index.html
```diff
@@ -112,7 +112,7 @@
       </div>
 
       <div class="flex items-center">
-        <a href="http://builder.openzeppelin.com" target="_blank" rel="noopener noreferrer" class="bg-[#3d3b5d] text-white px-4 py-2 rounded-full text-sm font-medium flex items-center gap-2 hover:bg-[#4a4872] transition-colors no-underline whitespace-nowrap">
+        <a href="https://builder.openzeppelin.com?ecosystem=evm" target="_blank" rel="noopener noreferrer" class="bg-[#3d3b5d] text-white px-4 py-2 rounded-full text-sm font-medium flex items-center gap-2 hover:bg-[#4a4872] transition-colors no-underline whitespace-nowrap">
```
ericglau: What does the ?ecosystem=evm parameter do?

packages/ui/public/stellar.html
```diff
@@ -62,12 +62,30 @@
...
+        <div>After deploying a contract, use the</div>
+        <div>new UI Builder to spin up a quick UI</div>
+      </div>
+
+      <div class="flex items-center">
+        <img src="/icons/arrow-right.svg" alt="arrow-right" class="arrow-gray">
+      </div>
+
+      <div class="flex items-center">
+        <a href="https://builder.openzeppelin.com?ecosystem=stellar" target="_blank" rel="noopener noreferrer" class="bg-[#3d3b5d] text-white px-4 py-2 rounded-full text-sm font-medium flex items-center gap-2 hover:bg-[#4a4872] transition-colors no-underline whitespace-nowrap">
```
ericglau: What does the ?ecosystem=stellar parameter do? I don't see any difference with or without it in UI builder.

packages/ui/public/stellar.html
```diff
@@ -62,12 +62,30 @@
     </div>
   </div>
   <div class="nav-row flex flex-row items-center justify-between">
-    <div class="nav">
+    <div class="nav nav-switches flex flex-row">
       <a class="switch switch-solidity" href="/"><img src="/icons/solidity.svg" alt="solidity">Solidity</a>
       <a class="switch switch-cairo" href="/cairo"><img src="/icons/cairo.svg" alt="cairo">Cairo</a>
-      <a class="switch switch-stellar active" href="#"><img src="/icons/stellar.svg" alt="stellar">Stellar</a>
+      <a class="switch switch-stellar active" href="/stellar"><img src="/icons/stellar.svg" alt="stellar">Stellar</a>
```
ericglau: If this is `/stellar`, the clicking Stellar while on Stellar tab causes the page to refresh, which seems pointless.  Using `#` will cause it to do nothing, which is preferable.
```suggestion
      <a class="switch switch-stellar active" href="#"><img src="/icons/stellar.svg" alt="stellar">Stellar</a>
```

packages/ui/public/index.html
ericglau: For EVM, would it look cleaner to omit `ecosystem` parameter if that is the default?

PR #791, opened 2026-04-01
Add trusted publishing

ericglau:
## Summary

Add npm trusted publishing support to the publish workflow while keeping `NPM_TOKEN` fallback for first-time publishes of new packages.

### Changes

- Add `registry-url: 'https://registry.npmjs.org'` to the shared setup action so `setup-node` writes npm config for the npm registry and publish auth is wired correctly.
- Add `id-token: write` to `.github/workflows/publish.yml` so GitHub Actions can mint the OIDC token required by npm trusted publishing.
- Install `npm@11.12.1` in the publish workflow because trusted publishing requires npm `11.5.1+`, while the default npm on Node 22 is older.
- Switch publish auth from `NPM_TOKEN` to `NODE_AUTH_TOKEN` because `setup-node` writes `.npmrc` using `${NODE_AUTH_TOKEN}`, which is the env var npm will actually read.
- Add `NPM_CONFIG_PROVENANCE: true` so provenance is generated consistently, including when the workflow falls back to token-based publish.

### Notes

- Keeps `NPM_TOKEN` fallback for new packages that do not yet have a trusted publisher configured.


tirumerla (approved):
Not sure if we still need the token, otherwise lgtm

tirumerla (approved):

pasevin (approved):
lgtm

PR #794, opened 2026-04-08
Use Node 24 for publish and fix OIDC auth handling

ericglau:
- Use Node 24 (which ships with npm 11) instead of installing npm 11 globally, fixing a self-upgrade MODULE_NOT_FOUND failure.
- Explicitly clear `NODE_AUTH_TOKEN` on the OIDC path to prevent setup-node's injected token from blocking trusted publishing.
- Parameterize the setup action's node-version and bump setup-node to v6.3.0.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,11 +1,16 @@
 name: Setup
 
+inputs:
+  node-version:
+    description: Node.js version to use
+    default: '22'
+
 runs:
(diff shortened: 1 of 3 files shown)


tirumerla (approved):

ericglau:
@SocketSecurity ignore-all
Looks like AI false positives

son-oz (approved):
LGTM

PR #797, opened 2026-04-23
Fix OIDC publish by stripping _authToken from .npmrc

ericglau:
Fixes the release workflow's OIDC trusted publishing path so it can authenticate without an `NPM_TOKEN`.

`actions/setup-node` writes `_authToken=${NODE_AUTH_TOKEN}` into `.npmrc`. In OIDC mode that variable is unset, and yarn errors when it tries to substitute it. This strips the `_authToken` line before yarn runs, leaving a clean config for OIDC to work with.


tirumerla (approved):

son-oz (approved):
LGTM

PR #798, opened 2026-04-28
Cairo: Split validation workflows

immrsd:
Splits the overloaded CI workflows that validate the generated Cairo and Cairo-alpha test projects using matrixes


immrsd:
@ericglau Once the PR gets merged, could you add the newly added workflows as mandatory checks for Cairo and Cairo-alpha?

.github/workflows/compile-cairo-alpha-project.yml
```diff
@@ -46,33 +46,13 @@ jobs:
           # Exit immediately if a command exits with a non-zero status
           set -e
 
-          declare -a all_kinds=("ERC20" "ERC721" "ERC1155" "ERC6909" "Account" "Multisig" "Governor" "Vesting" "Custom")
+          declare -a all_kinds=("ERC20" "ERC6909" "Account" "Multisig" "Governor" "Vesting" "Custom")
```
bidzyyys: Shouldn't we also extract `ERC20`?

.github/workflows/compile-cairo-project.yml
```diff
@@ -46,30 +46,13 @@ jobs:
           # Exit immediately if a command exits with a non-zero status
           set -e
 
-          declare -a all_kinds=("ERC20" "ERC721" "ERC1155" "Account" "Multisig" "Governor" "Vesting" "Custom")
+          declare -a all_kinds=("ERC20" "Account" "Multisig" "Governor" "Vesting" "Custom")
```
bidzyyys: Shouldn't we extract ERC20?

bidzyyys (commented):
@immrsd can't we just have a single job definition (easier to maintain) and then just invoke it for many different contracts matrices?

immrsd:
Refactored to leverage matrix

.github/workflows/compile-cairo-alpha-project.yml
immrsd: Now each type runs separate using matrix

ericnordelo (approved):
The workflow looks good, but it would be great to see it passing on this PR.

immrsd:
> The workflow looks good, but it would be great to see it passing on this PR.

None of the files in Cairo and Cairo-alpha packages were touched so the workflow isn't triggered. We also have to modify the repository's settings to require `validate-cairo-summary` job to succeed instead of `validate-cairo`, and same for `validate-cairo-alpha-summary`

The check will run on the PR with ERC-6909 extensions once this PR has been merged

immrsd:
@ericglau Could you take a look at the PR?

ericglau (approved):
LGTM, thanks

PR #806, opened 2026-05-27
Add TRON support to Contracts Wizard

km631:
## Summary

Adds TRON support to Contracts Wizard, targeting `@openzeppelin/tron-contracts` and `@openzeppelin/tron-contracts-upgradeable` `5.6.0-rc.2` through a structured print profile.

- New `/tron` Wizard surface for TRC20, TRC721, TRC1155, Governor, and Custom contracts
- TRON-specific names/imports applied through `tronPrintProfile` while printing library symbols and paths (never by rewriting finished source or user values)
- Solidity capped at `0.8.26`, the current `tron-solc` maximum
- Hardhat TRON and TronBox project downloads, including manual UUPS and transparent proxy deployment scaffolding
- Interactive MCP Apps for all five TRON tools
- CLI, MCP, AI assistant, Remix links/remappings, navigation, and changeset wiring

## Supported scope

- Account is omitted because ERC-4337 EntryPoint support is out of scope on TRON.
- Stablecoin and RealWorldAsset are omitted because their Community Contracts dependencies are not ported to TRON.
- OP Stack superchain bridging is omitted.
- ERC721/ERC1155 native bridging and Governor cross-chain execution are omitted because they are Contracts 5.7 APIs not present in TRON Contracts 5.6.
- Namespaced storage annotations intentionally remain `erc7201:` to match the published upgradeable library.

## Upgradeability

UUPS and transparent generation are enabled. Generated contracts import `@openzeppelin/tron-contracts-upgradeable`; downloadable projects compile a `TRC1967Proxy` or `TransparentUpgradeableProxy` helper and deploy the implementation and proxy directly. They do not use the EVM Hardhat Upgrades deployment flow.

## Verification

- Generated source compile test covers all supported option combinations, including upgradeable contracts, against the published rc.2 packages.
- Hardhat TRON and TronBox downloads are snapshot-tested.
- Both zip formats run `npm install` and compile for a basic project and a UUPS-upgradeable project.
- TypeScript preparation, formatting, and lint checks pass.
- No TRE/Docker deployment is run in the standard suite; that can be considered separately.


km631:
## Added MCP + CLI surface (commit 83046f7)

Followed up with `tron-*` tools for the MCP server and CLI registry so AI agents and CLI users get TRC-correct source out of the box (no manual `ERC* -> TRC*` rewrite needed when targeting TRON).

**MCP tools (7):**
- `tron-trc20`, `tron-trc721`, `tron-trc1155`
- `tron-stablecoin`, `tron-rwa`, `tron-governor`, `tron-custom`

**CLI commands (7, same names).** Try locally:
```
npx contracts-cli tron-trc20 --name MyToken --symbol MTK
```

**Design notes:**
- Schemas are reused verbatim from `solidityERC20Schema`/`solidityERC721Schema`/etc. — TRON options are identical to Solidity, only the source-level rendering differs. If TRON ever needs option-shape divergence, the schemas can fork at that point without touching the existing solidity ones.
- TRON prompts (e.g. _"Make a fungible token per the TRC-20 standard, targeting the TRON Virtual Machine."_) live in `packages/common/src/ai/descriptions/tron.ts` so they're shared across MCP + CLI.
- `Account` is intentionally excluded from the TRON tool set (same as the UI — ERC-4337 EntryPoint is out of scope).
- Upgradeable options are still accepted in the schemas; the upgradeables-pending caveat in the PR description applies to those imports regardless of whether they come from the UI download or an MCP/CLI tool.

**Tests:** 15 new MCP tests (78 total passing) + 12 new CLI tests (89 total passing). Lint clean.

packages/ui/src/tron/App.svelte
```diff
@@ -0,0 +1,72 @@
...
+  // Uses the Solidity Wizard with overrides specific to TRON:
+  //  - rewrite generated source to TRC token names + @openzeppelin/tron-contracts paths
+  //  - swap Hardhat download for the @openzeppelin/hardhat-tron-based one
+  //  - replace the second download tab (originally Foundry) with TronBox
+  //  - hide upgradeable downloads (tron-contracts upgradeable variants not yet shipped)
+  //  - remap @openzeppelin/tron-contracts when opening in Remix (TVM is EVM-compatible enough)
+  //  - hide Account tab (ERC-4337 EntryPoint not deployed on TRON in scope here)
+  const overrides: Overrides = {
+    omitTabs: ['Account'],
+    tabLabels: { ERC20: 'TRC20', ERC721: 'TRC721', ERC1155: 'TRC1155' },
```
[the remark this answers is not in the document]
km631: Skipping this one — the wizard does not have an `ERC4626` `Kind`/tab. The tabs are `ERC20 / ERC721 / ERC1155 / Stablecoin / RealWorldAsset / Account / Governor / Custom` (see [`Kind`](https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/core/solidity/src/kind.ts)), so `tabLabels` has no `ERC4626` key to remap. `ERC4626` references only appear as symbols/imports inside generated source (e.g. a tokenized-vault `Custom` contract), and those are already rewritten to `TRC4626` + `@openzeppelin/tron-contracts/token/TRC20/extensions/TRC4626.sol` by `rewriteForTron`'s symbol pattern.

ericglau:
Thanks for the PR!  Some initial comments:
- We should remove the Upgradeability options for now (or hide them in the UI and use a low-touch approach to disable the options in the API/CLI/MCP), otherwise it results in using `@openzeppelin/contracts-upgradeable` which won't compile together with the Tron contracts.
- Should we remove Stablecoin and RWA?  Those require ERC20Freezable and ERC20Restricted from `@openzeppelin/community-contracts` so those do not work here either, unless they are added to the Tron contracts.  cc @luiz-lvj
- According to [Tron docs](https://developers.tron.network/docs/concensus) it appears block time is 3 seconds.  There are some calculations that use block time, particularly Governor's Token Clock Mode if using Block Number, where it assumes 1 block is 12 seconds.  This is configurable in the user-facing options, but we should update this default for Tron.

ericglau:
### Consider replacing `rewriteForTron` (regex over rendered source) with a structured print transform

`rewriteForTron` rewrites the *final Solidity text*, so it can't distinguish what should be renamed (OZ library symbols + `@openzeppelin/contracts` import paths) from what shouldn't (user-supplied strings and the user's own identifiers). Because it substitutes globally over the whole string, real inputs get corrupted:

- **Name/symbol literals:** `\bERC(20|721|1155|4626)` matches inside constructor string literals. Symbol `"ERC20"` or name `"My ERC20 Token"` → `"TRC20"` / `"My TRC20 Token"`, silently changing the deployed `name()`/`symbol()`. The generated Hardhat/TronBox test asserts the *original* `opts.name`, so it fails by construction when the name contains a standard token.
- **Contract name:** a contract named `ERC20Token` → `contract TRC20Token`, but the zip still writes `contracts/ERC20Token.sol` and emits `artifacts.require('./ERC20Token.sol')` / `getContractFactory("ERC20Token")` → name/artifact mismatch, won't compile.
- **`securityContact`:** `ERC20-team@x.com` → `TRC20-team@x.com`.
- **Pragma cap** only matches `pragma solidity ^0.8.NN;` and only clamps the minor; any change to the emitted pragma shape silently ships an uncompilable version.

(Aside, non-blocking: the regex also renames `ERC20`→`TRC20` inside *other* roots like `@openzeppelin/community-contracts/` and `-upgradeable/`. The direction may be what you'd eventually want if TRC-flavored counterparts ship, so this isn't a bug to fix now — it just can't be decided correctly by one blanket rule, and overlaps the deferred library/upgradeable items. Another symptom of operating on text rather than structure.)

A structured transform avoids the corruption above because the generator already has the information the regex is missing. `printContract(contract, opts)` renders the structured `Contract` IR through two helpers — `transformName` (every inherited symbol: inheritance list, override lists, arg types, `using` libs) and `transformImport` (every import path). This is exactly how `@openzeppelin/contracts-upgradeable` output is produced today: `upgradeableName`/`upgradeableImport` in `options.ts` rename `ERC20`→`ERC20Upgradeable` and rewrite the path root. `Options.transformImport` is already a public hook.

Suggested shape — define a TRON "library profile" `Options` and drop `rewriteForTron`:
- `transformImport`: rewrite the `@openzeppelin/contracts/` root → `@openzeppelin/tron-contracts/` and rename `token/ERC20|721|1155` + symbol to `TRC*`. Because it sees each import's full path, it can make the right *per-package* decision — leave `community-contracts`/`contracts-upgradeable` as-is now, or map them to TRC equivalents later — instead of one blanket regex rule.
- add a `transformName` hook in `Options`/`withHelpers` (today only `transformImport` is exposed) that renames only OZ token-standard symbols.
- carry the capped solc version in the profile and parameterize the `pragma` line, instead of the pragma regex (and source the hardhat/tronbox config versions from the same constant).
- use the existing `additionalCompatibleLibraries` to emit `// Compatible with OpenZeppelin TRON Contracts …` (also closes the deferred header follow-up).

Then route every surface — UI display, zip generators, CLI/MCP — through that profile instead of `rewriteForTron`, collapsing five `rewriteForTron(...)` sites into one shared profile. `*.print(opts)` doesn't accept a print-`Options` today, but there's no need to widen it: `printContract` and `buildGeneric` are both exported from `@openzeppelin/wizard`, so a surface can call, for example, `printContract(buildGeneric({ kind, ...opts }), tronProfile)` in place of `rewriteForTron(erc20.print(opts))` (the zip generators already hold a built `Contract`, so they just pass the profile: `printContract(c, tronProfile)`). Because `transformName`/`transformImport` only see structured `ReferencedContract`/`ImportContract` nodes, user `name`/`symbol` (rendered by `printValue`), the contract name (printed directly), and `securityContact` (a NatSpec tag) are never touched. It's also more correct: a user contract named `ERC20Token` stays `ERC20Token` while only the inherited bases become `TRC*`.

This resolves the rewrite-rooted issues only; a few feature-gating, packaging, and consistency items are noted inline.

packages/cli/src/registry.ts
```diff
@@ -148,4 +159,37 @@ export const registry = {
 
   // Uniswap Hooks
   'uniswap-hooks': createRegistryEntry(uniswapHooksHooksSchema, opts => hooks.print(opts), uniswapHooksPrompts.Hooks),
+
+  // TRON (Solidity post-processed for @openzeppelin/tron-contracts + TRC* token names)
+  'tron-trc20': createRegistryEntry(solidityERC20Schema, opts => rewriteForTron(erc20.print(opts)), tronPrompts.TRC20),
```
ericglau: `superchain` cross-chain bridging is sanitized away in the UI (`tron/handle-unsupported-features.ts`: `defineOmitFeatures` hides it, `sanitizeOmittedFeatures` rewrites `superchain`→`custom`), but these CLI entries and the MCP `tron-*` tools reuse the full `solidity*Schema` and pass `crossChainBridging` straight into `print` unsanitized. `tron-trc20 --crossChainBridging superchain` (and the MCP equivalents for trc20/stablecoin/rwa) emit `@openzeppelin/tron-contracts/token/TRC20/extensions/draft-TRC20Bridgeable.sol` (doesn't exist) plus the hardcoded OP-Stack predeploy `0x4200…0028`, which is meaningless on TVM.

This isn't a deviation from Polkadot: Polkadot uses the identical UI sanitizer and avoids the leak only because it has no CLI/MCP surfaces — TRON is the first restricted variant to add those, so it's the first that needs the gate at that layer (mainline Solidity intentionally allows `superchain`, so there's no server-side precedent to copy).

Suggestion: promote `sanitizeOmittedFeatures` to a shared helper (`@openzeppelin/wizard` or `wizard-common`) and apply it in the `tron-*` registry entries and MCP tool handlers so all three surfaces share one gating definition — or disallow `superchain` for TRON at the schema level.

ericglau (commented):
Inline notes from a focused review of the TRON variant — correctness plus consistency with the patterns established by the other Wizard languages. The larger `rewriteForTron` → structured-transform suggestion is in a separate top-level comment.

packages/core/solidity/src/environments/hardhat/tron/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "hardhat test"
+  },
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@nomicfoundation/hardhat-chai-matchers": "^2.0.6",
+    "@nomicfoundation/hardhat-ethers": "^3.0.9",
+    "@openzeppelin/hardhat-tron": "^1.0.0",
```
ericglau: These version ranges can't resolve to the versions the PR description says will be published (`hardhat-tron@0.1.0`, `tron-contracts@0.0.1`): `^1.0.0` excludes all `0.x`, and `^0.1.0` (line 15) excludes `0.0.1`. So `npm install` in the generated zip will fail to resolve even *after* the packages publish, which contradicts the "no code changes required when they publish" note. Align the ranges with the intended publish versions (e.g. `@openzeppelin/hardhat-tron: ^0.1.0`, `@openzeppelin/tron-contracts: ^0.0.1`), and apply the same to the `tron-contracts` pin in `zip-tronbox.ts` (`'@openzeppelin/tron-contracts': '^0.1.0'`).

packages/core/solidity/src/zip-tronbox.ts
```diff
@@ -0,0 +1,286 @@
...
+  // above the deploy call so the user knows what to fill in.
+  return arg.name;
+}
+
+function declareArgPlaceholders(c: Contract): string[] {
+  return c.constructorArgs.map(arg => {
+    if (arg.type === 'address') {
+      return `// TODO: Replace with a real address (e.g. tronWeb.defaultAddress.base58).\n  const ${arg.name} = '<TRON address>';`;
+    }
+    return `// TODO: Set value for the ${arg.name} constructor argument.\n  const ${arg.name} = undefined;`;
```
ericglau: This emits `const <name> = undefined;` for non-address constructor args and then calls `deployer.deploy(<Name>, <name>)` unconditionally, so an unedited `tronbox migrate` compiles and silently deploys with `undefined` (e.g. a Governor's `IVotes` token).

Mainline `zip-hardhat`/`zip-foundry` never do this: address args get a usable value (`(await ethers.getSigners())[i].address` / `vm.addr(i)`); non-address args are emitted as *commented-out* declarations (`// const <name> = ...;`); and when any arg is unset the whole deploy/instance line is commented out behind a `// TODO: set the args, then uncomment`, so nothing runs until the user fills them in. (Foundry's script also uses an invalid sentinel like `<Set X address here>` that won't compile.)

Match that pattern here — no runtime guard, just placeholders that fail loudly: drop the `= undefined` assignment and either comment out the `deployer.deploy(...)` call (with a TODO to fill + uncomment) or use an obviously-invalid sentinel like the `'<TRON address>'` you already emit for address args. An unedited migration should error or no-op, never deploy `undefined`.

packages/core/solidity/src/zip-hardhat-tron.ts
```diff
@@ -0,0 +1,167 @@
...
+
+\`\`\`
+npx hardhat run --network tre scripts/deploy.ts
+\`\`\`
+
+The default \`tre\` network in \`hardhat.config.ts\` points at a local TRON Runtime Environment (Docker container, spawned automatically by \`@openzeppelin/hardhat-tron\`). For Shasta, Nile, or mainnet, add a network entry and pass \`--network <name>\`.
+`;
+  }
+
+  override async zipHardhat(c: Contract, opts?: GenericOptions): Promise<JSZip> {
```
ericglau, line 288: `override` is used here on `zipHardhat` but omitted on the other ~11 methods in this class that also override `HardhatZipGenerator` (the sibling `zip-hardhat-polkadot.ts` uses it on none). It compiles today only because `noImplicitOverride` isn't enabled. Suggest making it consistent — add `override` to all the overriding methods (preferred) or none — so it won't break if that compiler flag is ever turned on.

packages/ui/src/tron/App.svelte
```diff
@@ -0,0 +1,78 @@
...
+  };
+</script>
+
+<div class="tron-app">
+  <SolidityApp {initialTab} {initialOpts} {overrides} on:tab-change={event => dispatch('tab-change', event.detail)} />
+</div>
+
+<style lang="postcss">
+  .tron-app :global(.tab button.selected) {
+    background-color: #ff060a !important;
```
ericglau: Every other language defines its brand color as a CSS variable in `packages/ui/src/common/styles/vars.css` (e.g. `--polkadot-pink`) and references `var(--…)`; TRON hardcodes `#ff060a` inline here (and again in `standalone.css`). Suggest adding `--tron-red: #ff060a;` to `vars.css` and referencing it in both places to match the pattern and avoid the duplicated literal.

packages/common/src/ai/descriptions/solidity.ts
```diff
@@ -85,7 +85,7 @@ export const solidityAccountDescriptions = {
 export const solidityGovernorDescriptions = {
   delay: 'The delay since proposal is created until voting starts, default is "1 day"',
   period: 'The length of period during which people can cast their vote, default is "1 week"',
-  blockTime: 'The block time of the chain in seconds, default is 12',
+  blockTime: 'The block time of the chain in seconds. Defaults to 12 for Solidity (Ethereum) and 3 for TRON.',
```
ericglau: This shared Solidity description now mentions TRON, so the TRON-specific note leaks into the Solidity (and Polkadot) `governor` `--help` and the `solidity-governor` MCP tool schema — everywhere `solidityGovernorSchema` is consumed (it's the only consumer of this string; the UI control uses `governor.defaults` instead). It would be cleaner to have them separated.

Since `solidityGovernorSchema` is a raw `z.ZodRawShape`, you can keep this description Solidity-specific and override just the one field for TRON:

1. Revert this line to `'The block time of the chain in seconds, default is 12'`.
2. Add a TRON description in `descriptions/tron.ts` (its no-imports rule is fine — plain string):
   ```ts
   export const tronGovernorDescriptions = {
     blockTime: 'The block time of the chain in seconds, default is 3.',
   };
   ```
3. Add `schemas/tron.ts` (re-export from the schemas barrel), overriding only `blockTime`:
   ```ts
   export const tronGovernorSchema = {
     ...solidityGovernorSchema,
     blockTime: z.number().optional().describe(tronGovernorDescriptions.blockTime),
   } as const satisfies z.ZodRawShape;
   ```
4. Point `tron/tools/governor.ts` and the `tron-governor` registry entry at `tronGovernorSchema`.

The default value (3) is still applied at the call site via `blockTime ?? TRON_DEFAULT_BLOCK_TIME`, so this is description-only — no behavior change.

packages/cli/src/registry.ts
km631: Done in 5b9aba8. Promoted the gate to a shared `sanitizeTronOptions` helper in `@openzeppelin/wizard` and applied it in the `tron-trc20`/`-stablecoin`/`-rwa` CLI registry entries and the matching MCP handlers; the UI's `sanitizeOmittedFeatures` now delegates to the same helper, so all three surfaces share one definition. `superchain` is downgraded to `custom` before printing — verified the TRON output no longer contains the OP-Stack `0x42…0028` predeploy, while mainline Solidity is unchanged.

packages/core/solidity/src/environments/hardhat/tron/package.json
km631: Done in 5b9aba8. Aligned with the intended publish versions: `@openzeppelin/hardhat-tron: ^0.1.0` and `@openzeppelin/tron-contracts: ^0.0.1` here, and the same `@openzeppelin/tron-contracts: ^0.0.1` pin in `zip-tronbox.ts`. (Confirmed against the source repos: hardhat-tron is at 0.1.0, tron-contracts at 0.0.1.)

packages/core/solidity/src/zip-tronbox.ts
km631: Done in 5b9aba8. Dropped the `= undefined` assignment. Non-address args now emit commented-out declarations, and when any arg is unset the whole `deployer.deploy(...)` call is commented out behind a `// TODO: Uncomment once the constructor arguments above are set.` — mirroring mainline `zip-hardhat`/`zip-foundry`. Address args keep the obviously-invalid `'<TRON address>'` sentinel. So an unedited `tronbox migrate` either fails loudly (invalid address) or no-ops, never deploys `undefined`.

packages/core/solidity/src/zip-hardhat-tron.ts
km631: Done in 5b9aba8. Went with the preferred option — added `override` to all ~10 methods of `HardhatTronZipGenerator` that override `HardhatZipGenerator`, so it stays correct if `noImplicitOverride` is ever enabled.

packages/ui/src/tron/App.svelte
km631: Done in 5b9aba8. Added `--tron-red: #ff060a;` to `vars.css` and referenced `var(--tron-red)` from both `App.svelte` and `standalone.css`.

packages/common/src/ai/descriptions/solidity.ts
km631: Done in 5b9aba8, following your suggestion. Reverted this to `'The block time of the chain in seconds, default is 12'`, added `tronGovernorDescriptions.blockTime` (default 3) in `descriptions/tron.ts` and a `tronGovernorSchema` overriding only `blockTime` in `schemas/tron.ts`, and pointed both the `tron-governor` CLI registry entry and the MCP tool at it. The default value is still applied at the call site via `blockTime ?? TRON_DEFAULT_BLOCK_TIME`, so this is description-only. Verified the `solidity-governor` `--help` now reads "default is 12" while `tron-governor` reads "default is 3".

ericglau:
Thanks, this looks ready from the items I flagged:

- [x] Enable the TRON compile test. `packages/core/solidity/src/tron-compile.test.ts` compiles every supported TRON option combination, including upgradeable contracts, against `@openzeppelin/tron-contracts(-upgradeable)@5.6.0-rc.2`. This catches ERC→TRC import/symbol mismatches that content snapshots cannot.
- [x] Verify download zips beyond snapshots: snapshots + `npm install && compile`, no TRE. Both Hardhat TRON and TronBox cover basic non-upgradeable and UUPS-upgradeable projects. A full Docker/TRE deploy remains an optional follow-up, not part of this PR.
- [x] Resolve the ERC→TRC alias boundary: all ERC/EIP library symbols map to TRC/TIP.
- [x] Keep `@custom:storage-location erc7201:<id>` (not `trc7201:`). This is intentional in tron-contracts-upgradeable; Wizard matches it for consistency. The slot formula is unchanged; only the NatSpec label differs.

ericglau:
@SocketSecurity ignore-all
Looks like false positives.

ericglau:
@SocketSecurity ignore-all

ericglau (approved):

PR #810, opened 2026-06-08
Update Hardhat sample project download to Hardhat 3

ericglau:
## Summary

Updates the standard Solidity wizard **Download Hardhat project** output from Hardhat 2 to Hardhat 3.

- Generated projects are ESM and use Hardhat 3, `@nomicfoundation/hardhat-ethers` v4, and Hardhat Ignition v3 (if non-upgradeable)
- Upgradeable downloads use `@openzeppelin/hardhat-upgrades` v4 which is for Hardhat 3
- Tests use AVA + tsx; deploy/test code uses `hre.network.create()`
- `HardhatZipGenerator` (Hardhat 2) remains the shared base class; Polkadot and Confidential extend it for plugin-specific overrides while the standard Solidity download uses `Hardhat3ZipGenerator`. Those variants stay on Hardhat 2 because `@parity/hardhat-polkadot` and `@fhevm/hardhat-plugin` do not yet support Hardhat 3

Does not change the wizard's own Hardhat version for development (still HH2 for compile tests and `prepare.ts`). That internal tooling uses HH2's programmatic API and custom compile-task hooks; migrating it is a separate effort unrelated to the self-contained sample projects this PR generates.

## Test plan

Covered by `packages/core/solidity/src/zip-hardhat.test.ts`: extracts zips, runs `npm ci && npm test` (`hardhat compile && ava`), and deploys for both Ignition and upgradeable paths. Polkadot and Confidential are unchanged.


packages/core/solidity/src/environments/hardhat/package.json
```diff
@@ -1,16 +1,24 @@
...
-    "hardhat": "^2.22.0"
+    "ava": "^6.0.0",
+    "ethers": "^6.8.1",
+    "hardhat": "^3.6.0",
+    "tsx": "^4.7.0",
+    "typescript": "^5.0.0"
   }
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
ericglau: Disagree. We should not enforce Node versions for Hardhat -- Hardhat gives useful error if using an unsupported Node version, and Node versions below v22 are all EOL anyways.

pasevin (approved):

packages/core/solidity/src/zip-hardhat.ts
```diff
@@ -308,6 +315,246 @@ ${c.upgradeable ? 'npx hardhat run --network <network-name> scripts/deploy.ts' :
...
+  private getImports(c: Contract): Lines[] {
+    const imports = ['import test from "ava";', 'import hre from "hardhat";'];
+    if (c.upgradeable) {
+      imports.push('import { upgrades } from "@openzeppelin/hardhat-upgrades";');
+    }
+    return imports;
+  }
+
+  private getConnectionSetup(c: Contract): Lines[] {
+    const lines = ['const connection = await hre.network.create();', 'const { ethers } = connection as any;'];
```
ernestognw: The tests declare `const { ethers } = connection as any;` but the deploy script declares `const { ethers } = connection;`. I noticed the asymmetry but I'm not sure what's the reasoning; I understand the cast is needed because the test file doesn't import any plugin that augments `NetworkConnection` with `ethers`.

If I'm right, importing `@openzeppelin/hardhat-upgrades` should pull the augmentation

packages/core/solidity/src/zip-hardhat.ts
```diff
@@ -308,6 +315,246 @@ ${c.upgradeable ? 'npx hardhat run --network <network-name> scripts/deploy.ts' :
...
+      ? `ContractFactory.deploy(${args.join(', ')})`
+      : unsafeAllowConstructor
+        ? `upgradesApi.deployProxy(ContractFactory, [${args.join(', ')}], { unsafeAllow: ['constructor'] })`
+        : `upgradesApi.deployProxy(ContractFactory, [${args.join(', ')}])`;
+  }
+
+  protected getScript(c: Contract): string {
+    return `\
+import hre from "hardhat";
+import { upgrades } from "@openzeppelin/hardhat-upgrades";
```
ernestognw, line 488: This `getScript` imports `@openzeppelin/hardhat-upgrades` unconditionally, which seems to be fine because this function is only called in `getPrintContract` only if `c.upgradeable`. However, I noticed the `HardhatZipGenerator`'s `getScript` is sensitive to `c.upgradeable`. This feels inconsistent, but I can't point out to exactly what's wrong. I believe it's worth a second look

ernestognw (approved):
Not blocking imo, but worth taking a look to the following comments:

packages/core/solidity/src/zip-hardhat.ts
ericglau: Good catch — the asymmetry pointed at a real gap. Importing `@openzeppelin/hardhat-upgrades` did not pull the augmentation in 4.0.1: the plugin loads hardhat-ethers at runtime but its types didn't reference it, so the deploy script actually failed `tsc --noEmit` and the test only passed because of the cast. This is fixed upstream in hardhat-upgrades 4.0.2 (OpenZeppelin/openzeppelin-upgrades#1275), which re-exports the hardhat-ethers types. Updated this PR to use `^4.0.2` and removed the cast, so the test and script now use the same declaration.

packages/core/solidity/src/zip-hardhat.ts
ericglau: Good instinct. The conditional in the Hardhat 2 `getScript` predates Ignition support, when deploy scripts were generated for both paths — since then, scripts are only generated for upgradeable contracts (`zipHardhat` routes non-upgradeable to an Ignition module), so that branch is dead code rather than intentional flexibility. Rather than porting it into the Hardhat 3 generator, made the invariant explicit with a comment and a guard that throws if `getScript` is ever called for a non-upgradeable contract, so a future caller bug fails loudly instead of producing a broken project.

Issue #219, opened 2023-01-12
Highlight differences when changing features


ericglau:
When adding or removing a selected feature, it may not be obvious what parts of the resulting code has changed.  This is especially the case for Cairo because there is a lot of code that requires scrolling to view.

Enhance the UI so that differences are visibly highlighted when changing a feature.

For example, if the user starts with:
![Screen Shot 2023-01-12 at 4 56 55 PM](https://user-images.githubusercontent.com/29640130/212189790-a7efcf2d-ea64-41b2-a2ec-3ec4322d037d.png)
then enables Mintable like the below:
![Screen Shot 2023-01-12 at 4 57 48 PM](https://user-images.githubusercontent.com/29640130/212189955-4acc85f6-2ff2-4ce4-b0e4-a302b564d223.png)
, this should cause the following sections to be highlighted:
```
import "@openzeppelin/contracts/access/Ownable.sol";
```
and
```
    function mint(address to, uint256 amount) public onlyOwner {
        _mint(to, amount);
    }
```

kilnhast:
Hey @ericglau ! can I get assigned to this issue?

ericglau:
@kilnhast Thanks, feel free to contribute a PR

makiopen:
Hi @kilnhast thanks for looking into this issue! We were thinking that it would be a nice motion interaction highlighting the differences if we'd animate the typing and removal of the code eg. https://cngu.github.io/vue-typer/?ref=madewithvuejs.com

Let me know if I can help in any way!

PR #811, opened 2026-06-08
Highlight changed Solidity code lines

M3stark:
Refs #219.

This draft PR adds a lightweight visual highlight for Solidity code lines that change after feature toggles.

Notes:
- Tracks the previous generated Solidity source in the UI component.
- Wraps changed lines in a highlight span after syntax highlighting and hyperlink injection.
- Keeps the change scoped to the Solidity wizard first, matching the ERC20/Mintable example in the issue.

Validation:
- Not run locally in this environment.

### DIFF
--- a/packages/ui/src/solidity/App.svelte
+++ b/packages/ui/src/solidity/App.svelte
@@ -81,6 +81,56 @@
   let contract: Contract = new ContractBuilder(initialOpts.name ?? 'MyToken');
 
   let showCode = true;
+  const previousCodeForDiff = new Map<Kind, string>();
+  let highlightedLineNumbers = new Set<number>();
+
+  function changedLineNumbers(previousCode: string, nextCode: string) {
+    if (previousCode === '') {
+      return new Set<number>();
+    }
+    const previousLines = previousCode.split('\n');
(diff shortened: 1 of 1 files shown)


M3stark:
@coderabbitai review

ericglau:
Thanks for the PR.  There's an issue with spacing, as sometimes empty lines are inserted between every line.

But overall I think this should take a different approach in terms of design.  Currently the highlights remain visible long after making changes to the selected option, and it's not clear to the user what the highlights mean.  Yellow particularly looks like a warning and users may mistaken the highlights as code issues.

Instead, perhaps one of the following would work better, using a combination of transitions and timing:
- Crossing out and/or fading out the lines that are going to be deleted, and after a very short delay (e.g. few hundred milliseconds), replacing those with the new lines.
- Or using some form of green and/or red to briefly show what will be changed
- Or otherwise highlighting the changed lines (perhaps with green) briefly after the new code is applied

But in any case, the steady state (e.g. after a few hundred milliseconds) there should be no long-term highlights visible in the code.

Curious to get your thoughts on the above and whether you would like to make those changes.

PR #812, opened 2026-06-09
Add optional decimals to Solidity ERC20/Stablecoin/RWA, Confidential ERC7984, and Stellar Fungible/Stablecoin

ericglau:
Adds an optional `decimals` field, in the same way as the existing Cairo ERC20 decimals option:

- **Solidity** `erc20`, `stablecoin`, `realWorldAsset`: overrides `decimals()` when set to a non-default value. Defaults to 18.
- **Confidential** `erc7984`: overrides `decimals()` when set to a non-default value. Defaults to 6. Incompatible with the Wrappable extension, which derives its decimals from the underlying token — combining them throws an `OptionsError`, and the UI disables the Decimals field while Wrappable is enabled (restoring the previous value when re-enabled).
- **Stellar** `fungible`, `stablecoin`: sets the token's metadata decimals. Defaults to 7.

Premint amounts are scaled to the chosen number of decimals, with overflow validation against the contract's premint type. Also replaces the internal `toUint8`/`toUint256` validators with a generic `toUint` validator exported from the Solidity package, and adds `decimals` to the AI assistant schemas and MCP tools.


pasevin (approved):

ericglau:
The `build (stellar, compile)` failure is unrelated to this PR — it affects any PR touching the Stellar package and is fixed by #815. Will merge master into this branch after that lands.

arr00 (approved):

PR #818, opened 2026-06-17
Reject line terminators in info.securityContact and info.license

ericglau:
`info.securityContact` and `info.license` are printed into a single-line comment in the generated source (the `// SPDX-License-Identifier:` line and the security-contact comment). If a value contains a line break, it no longer stays on that one comment line.

`setInfo` now throws `OptionsError` when either field contains `\n`, `\r`, `U+2028`, or `U+2029` — the newline characters plus the two Unicode line separators that editors and some compilers also treat as line breaks — keeping these metadata fields to a single line. Applied across the Solidity, Cairo, Stellar, and Stylus generators (and inherited by Confidential and Uniswap Hooks via the Solidity core), with tests and a changeset.

Also surfaces the resulting validation error inline on the Security Contact and License fields in the UI, consistent with how other field errors are shown.

### DIFF
--- a/.changeset/quick-spiders-cheer.md
+++ b/.changeset/quick-spiders-cheer.md
@@ -5,4 +5,4 @@
 '@openzeppelin/wizard-stylus': patch
 ---
 
-Reject line terminators in `info.securityContact` and `info.license` to prevent breaking out of the generated comment lines.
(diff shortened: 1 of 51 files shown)


son-oz (approved):
LGTM

0xNeshi (approved):

PR #821, opened 2026-06-23
Stellar vault

### LINKED ISSUE #745: Add Stellar Vault Contract
Feature Request: 
Include Stellar `Vault` contract example in the UI Wizard, alongside the existing `Fungible`, `NonFungible` and `Stablecoin` contracts.

- Wizard UI: https://wizard.openzeppelin.com/stellar
- Example Implementation: https://github.com/OpenZeppelin/stellar-contracts/blob/main/examples/fungible-vault/src/contract.rs

ozgunozerk:
Adds `Vault` to Stellar 

Fixes #745


brozorec (approved):
LGTM 👍 

Only question I have, shouldn't we set the default offset > 0 ?

packages/core/stellar/src/vault.compile.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      name: 'MyVault',
+      symbol: 'MTK',
+      access: 'roles',
+    },
+    { snapshotResult: false },
+  ),
+);
+
+test.serial(
+  'compilation vault full - ownable',
```
ericglau, line 77: Should we add a custom decimalsOffset as part of the "full" testcase(s)?

packages/core/stellar/src/vault.test.ts
```diff
@@ -0,0 +1,133 @@
...
+testVault('vault explicit trait implementations', {
+  explicitImplementations: true,
+});
+
+testVault('vault pausable explicit trait implementations', {
+  pausable: true,
+  explicitImplementations: true,
+});
+
+testVault('vault full - ownable', {
```
ericglau, line 80: Should we add a custom decimalsOffset as part of the "full" testcase(s)?

packages/common/src/ai/descriptions/stellar.ts
```diff
@@ -38,6 +39,11 @@ export const stellarStablecoinDescriptions = {
   votes: 'Whether to enable vote checkpoints and delegation for governance.',
 };
 
+export const stellarVaultDescriptions = {
+  decimalsOffset:
+    'Virtual decimals offset added to the underlying asset decimals to derive the vault share decimals. Higher values increase resistance to inflation (donation) attacks. Must be between 0 and 10. Defaults to 0.',
```
ericglau: Since this says higher values are more resistant to inflation attacks, it may not be clear to readers why 0 is a sensible default — e.g., is 0 already safe on its own? It might help to clarify that, along with any tradeoffs of using a value larger than 0.

packages/ui/src/stellar/VaultControls.svelte
```diff
@@ -0,0 +1,74 @@
...
+      <span>Symbol</span>
+      <input bind:value={opts.symbol} use:error={errors?.symbol} />
+    </label>
+  </div>
+
+  <label class="labeled-input">
+    <span class="flex justify-between pr-2">
+      Decimals offset
+      <HelpTooltip>
+        Virtual offset added to the underlying asset's decimals to derive the vault share decimals. A higher value
```
ericglau: Similar to my other comment in the schema description, we should consider clarifying the tradeoffs of using 0 vs higher values.  Although this is a tooltip so we should be succinct.

ozgunozerk:
> shouldn't we set the default offset > 0 ?

@brozorec , I don't think so. This is from our library:

```
/// While not fully preventing the attack, analysis shows that the default
/// offset (0) makes it non-profitable even if an attacker is able to capture
/// value from multiple user deposits, as a result of the value being captured
/// by the virtual shares (out of the attacker's donation) matching the
/// attacker's expected gains. With a larger offset, the attack becomes orders
/// of magnitude more expensive than it is profitable.
```

And this:

> If the offset is 0, the attacker’s loss is at least equal to the user’s deposit.

is from: https://docs.openzeppelin.com/contracts/5.x/erc4626

ericglau (approved):
LGTM, thanks!
One additional comment: when offset is 0, would it be preferable to omit the line `Vault::set_decimals_offset(e, 0);` from the resulting contract, or is it still required?

ozgunozerk:
> One additional comment: when offset is 0, would it be preferable to omit the line Vault::set_decimals_offset(e, 0); from the resulting contract, or is it still required?

It is not required, though I think it is good to leave as is for discoverability.

ericglau:
@ozgunozerk Thanks. Just a note, this repo requires signed commits now so I've signed and force-pushed your commits.

PR #824, opened 2026-07-03
fix(stellar): assert on raw token name in generated rust tests

Assassin859:
### Description

This PR fixes a bug in the generated Soroban Rust tests where the name assertion fails for token names containing spaces or special characters. 

### Root Cause
- The `ContractBuilder` constructor converts a user-supplied name (e.g. `"My Token"`) to a valid Rust struct identifier (e.g. `"MyToken"`) and stores it as `c.name`.
- When asserting on the token's name, the generated test files (`printRustNameTest` and `printVaultRustTest`) were using `c.name` (the Rust struct identifier). 
- However, at runtime, `client.name()` returns the raw token name supplied to `Base::set_metadata(...)` (e.g. `"My Token"`).
- Consequently, the generated tests would fail whenever the token name contained spaces or non-identifier characters.

### Changes
- Updated `printRustNameTest` and `printVaultRustTest` inside `zip-shared.ts` to accept `tokenName: string` and use it in the assertion rather than `c.name`.
- Threaded `opts.name` (the raw user input) into the test printing functions within `zip-rust.ts`.
- Added a unit test validating that raw token names with spaces generate the correct assertion string while keeping the sanitized identifier for the struct.

### DIFF
--- a/packages/core/stellar/src/zip-rust.ts
+++ b/packages/core/stellar/src/zip-rust.ts
@@ -40,7 +40,7 @@ export const createRustZipEnvironment = (c: Contract, opts: GenericOptions) => {
   zip.file(`contracts/${contractName}/src/contract.rs`, removeCreateLevelAttributes(printContract(c)));
   zip.file(
     `contracts/${contractName}/src/test.rs`,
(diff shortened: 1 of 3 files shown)


Assassin859:

PR #825, opened 2026-07-17
Add cross-chain options for ERC721, ERC1155, and Governor (Contracts 5.7)

ericglau:
Adds Wizard support for the new cross-chain contracts in OpenZeppelin Contracts 5.7.

## New options

- **ERC721 / ERC1155**: `crossChainBridging` (ERC-7786 Native) embeds a bridge directly in the token via `ERC721Crosschain`/`ERC1155Crosschain` — outbound transfers burn, inbound transfers mint. Adds a `CrosschainLinked.Link[]` constructor argument and an access-controlled `setLink` function, with the `crossChainLinkAllowOverride` sub-option. Mirrors the existing ERC20 `erc7786native` option.
- **Governor**: `crossChainExecution` adds the `GovernorCrosschain` extension, letting passed proposals relay execution to other chains through ERC-7786 gateways. The `CrosschainRemoteExecutor` on target chains is deployed by the user (not generated); gateway and executor are per-proposal arguments to `relayCrosschain`.

Combining ERC721 auto-increment IDs with bridging is allowed but shows a warning (IDs increment independently per chain; mint on a single chain and link non-minting counterparts to avoid collisions that strand bridged tokens).

Covers core codegen, UI, AI assistant descriptions/schemas/function definitions, MCP tools, and CLI.

## Fixes

Upgradeable `crossChainBridging` combinations were previously excluded from the exhaustive compile matrix; they are now covered via reduced blueprints, which surfaced two existing compile errors in upgradeable ERC20 bridging variants (fixed here):
- `erc7786native` referenced the untranspiled `CrosschainLinked.Link[]` in the initializer.
- `custom` with managed access transpiled the `AuthorityUtils` library import to a nonexistent file.

Two fixes outside the cross-chain surface:
- **Account**: the `IERC4337` import path dropped its `draft-` prefix in Contracts 5.7, so generated account contracts import `interfaces/IERC4337.sol`.
- **UI**: the clickable import links resolved `@openzeppelin/contracts` from a hoisted copy in `node_modules` and pointed at v5.4.0; they now use the version the Wizard's output is generated against. Affects all Solidity tabs.

Additionally, downloaded Foundry and Hardhat projects now compile when constructors have non-address arguments (e.g. cross-chain links, or Governor's token and timelock): the generated test and script files declare typed placeholders and comment out the deployment with a TODO, instead of calling the constructor with missing arguments.

## Before merging

- [x] Contracts 5.7.0 final released
- [x] Re-pin `@openzeppelin/contracts` and `@openzeppelin/contracts-upgradeable` to `^5.7.0` (core devDependencies and both Hardhat environments) and regenerate lockfiles/snapshot via `prepare` and `update-env`
- [x] Update `compatibleContractsSemver` to `^5.7.0`
- [x] Confirm API docs anchors for `ERC721Crosschain`, `ERC1155Crosschain`, and `GovernorCrosschain` resolve — the anchors are not published yet; they land with OpenZeppelin/docs#226. The help links already use the correct URLs and currently resolve to the right page without scrolling to the anchor.


packages/core/solidity/src/zip-hardhat.test.ts.md
```diff
@@ -555,13 +555,278 @@ Generated by [AVA](https://avajs.dev).
...
+      }`,
+      `import { buildModule } from "@nomicfoundation/hardhat-ignition/modules";␊
+      ␊
+      export default buildModule("MyTokenModule", (m) => {␊
+      ␊
+        // TODO: Set values for the constructor arguments below␊
+        const myToken = m.contract("MyToken", [links, initialOwner]);␊
+      ␊
+        return { myToken };␊
+      });␊
```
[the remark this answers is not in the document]
ericglau: Thanks — this pointed at a real weakness in the placeholder story, though in a different file than flagged. Splitting the two:

**Foundry (fixed in this PR).** The generated test and script declared typed placeholders whose types were never imported — `CrosschainLinked.Link[] memory links` with no `import {CrosschainLinked}`, and likewise `IVotes` / `TimelockController` for Governor. Following the TODO exactly ("set values, then uncomment") could not compile. Fixed by emitting the imports for non-address constructor argument types, which also surfaced a second bug: upgradeable projects declared `CrosschainLinked.Link[]` while the initializer expects `CrosschainLinkedUpgradeable.Link[]`, so the declared type now goes through the same name transformation as the contract itself. Verified by extracting a generated project, uncommenting, filling the placeholder, and running `forge build` — previously a compile error, now exits 0.

**Hardhat Ignition (no change).** These identifiers are intentional fill-in-the-blank scaffolding, and the pattern is pre-existing on `master` for every contract kind — e.g. `m.contract("MyToken", [recipient, defaultAdmin, pauser, minter])` with the same TODO. Hardcoding constructor arguments inline is also the documented Hardhat Ignition pattern; the official [Getting Started](https://hardhat.org/ignition/docs/getting-started) example is `m.contract("Rocket", ["Saturn V"])`, with `m.getParameter()` offered for configurability rather than correctness.

It also already fails closed in a self-locating way. Deploying without editing gives:

```
  const myToken = m.contract("MyToken", [links, initialOwner]);
                                         ^
ReferenceError: links is not defined
    at ignition/modules/MyToken.ts:6:42
```

— the caret lands directly under the TODO that explains the fix. Nothing in the normal flow evaluates the module either: `npm test` is `hardhat compile && ava`, neither of which touches `ignition/`.

The Foundry case was different in kind: `<Set links here>` is syntactically invalid Solidity *and* the declared types weren't imported, so `forge build` failed and the documented "uncomment" step couldn't work. That's what we fixed.

@coderabbitai resolve

packages/ui/src/solidity/GovernorControls.svelte
```diff
@@ -255,6 +256,13 @@
   </div>
 </ExpandableToggleRadio>
 
+<ExpandableCheckbox
+  label="Cross-Chain Execution"
+  bind:checked={opts.crossChainExecution}
+  helpContent="Lets passed proposals relay execution to other chains through ERC-7786 gateways. Requires a CrosschainRemoteExecutor contract, controlled by this governor, deployed on each target chain."
+  helpLink="https://docs.openzeppelin.com/contracts/5.x/api/governance#GovernorCrosschain"
```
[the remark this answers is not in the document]
ericglau: Correct that the anchor doesn't resolve yet — the 5.7 API docs haven't been published. The same applies to the `ERC721Crosschain` and `ERC1155Crosschain` links on the other two token tabs.

The URL itself is already correct: the API docs generation for Contracts 5.7 adds a `GovernorCrosschain` section to `content/contracts/5.x/api/governance.mdx` using exactly this anchor format, and the docs repo's own internal links use the same `/contracts/5.x/api/governance#GovernorCrosschain` shape. Until that deploys, the link resolves to the right API page without scrolling to the section.

Keeping it as-is rather than dropping it, since removing a correct link and re-adding it shortly after is the worse trade. Tracked in the PR description checklist.

@coderabbitai resolve

packages/core/solidity/src/utils/version.ts
```diff
@@ -1,4 +1,4 @@
 /**
  * Semantic version string representing of the minimum compatible version of Contracts to display in output.
  */
-export const compatibleContractsSemver = '^5.6.0';
+export const compatibleContractsSemver = '^5.7.0';
```
ernestognw, line 4: In `inject-hyperlinks.ts`, we get the version from `openzeppelin-contracts-version.json`. Shall we do it here too?

packages/core/solidity/src/zip-shared.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Whether any constructor argument cannot be given a default value in a generated project.
+ *
+ * Address arguments are filled in with a signer, so a contract with only address arguments can be
+ * deployed as-is. Anything else must be supplied by the user, which means the generated deployment
+ * is commented out and there is no deployed instance for the generated tests to assert against.
+ */
+export function hasNonAddressArgs(c: Contract): boolean {
+  return c.constructorArgs.some(arg => arg.type !== 'address');
```
ernestognw:
```suggestion
  return c.constructorArgs.some(arg => arg.type !== 'address' && arg.type !== 'address payable');
```

packages/core/solidity/src/account.ts
ernestognw:
```suggestion
    // WebAuthnSigner inherits from P256Signer, so the AbstractSigner override is handled by `addSigner`
```

packages/ui/src/solidity/ERC721Controls.svelte
```diff
@@ -129,6 +159,34 @@
   </div>
 </ExpandableToggleRadio>
 
+<ExpandableToggleRadio
+  label="Cross-Chain Bridging"
```
ernestognw, line 163: We use "cross-chain" in some sections and "crosschain" and others. Shall we standardize to "crosschain"?

packages/ui/src/solidity/manual-notice-tooltip.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Tippy.js properties for a notice that is shown programmatically when the user selects a
+ * combination of options that needs a warning, rather than on hover.
+ */
+export function manualNoticeTooltipProps(content: string): { [prop: string]: string | boolean } {
```
ernestognw:
```suggestion
import type { Props } from 'tippy.js';

/**
 * Tippy.js properties for a notice that is shown programmatically when the user selects a
 * combination of options that needs a warning, rather than on hover.
 */
export function manualNoticeTooltipProps(content: string): Partial<Props> {
```

ernestognw (approved):
LGTM, few nits after a quick pass. Thanks @ericglau

packages/core/solidity/src/utils/version.ts
ericglau: This is intentionally kept separate. `openzeppelin-contracts-version.json` is the exact version we build and test against — it resolves to whatever the devDependency range installs, e.g. 5.7.1 from ^5.7.0 — whereas this semver describes the minimum version the generated code needs. Deriving one from the other would raise the stated minimum on every upstream patch release, and we may also want to bundle a newer version while the generated code stays compatible with an older one.

packages/ui/src/solidity/ERC721Controls.svelte
ericglau: Agreed, we should standardize on "crosschain". Opened #831 to do it across all the tabs, since the mix exists in ERC20's shipped text too and this PR adds more of it.

Note that renaming ERC20's existing `crossChain*` option keys in the API would be breaking, so we kept the new options aligned with that naming — the same concept spelled two ways across tabs is a footgun for anything consuming the options API, especially LLMs filling these in via the MCP tools and AI assistant. #831 suggests updating user-facing labels but leaves the keys as-is.

packages/core/solidity/src/zip-shared.ts
ericglau: Extracted the check into a shared `isAddressType()` in `zip-shared.ts` so the zip generators use one predicate instead of separate `=== 'address'` comparisons that had to stay in sync.

Left `address payable` out: no generator produces one today, unknown types already fail safe (the whole deployment goes behind the TODO, so the project still compiles), and supporting it properly would also need a `payable()` cast in the Foundry declaration since Solidity won't implicitly convert `address`. Easy to add on top of the shared helper if it comes up.

PR #826, opened 2026-07-28
Stellar: add account

brozorec:
Adds a Stellar `Account` contract type implementing the smart account framework, combining Delegated, Ed25519 and WebAuthn (passkey) signers with an optional threshold policy.

## What it generates

- Each enabled signer type contributes its own typed constructor arguments, which the constructor assembles into a single `Vec<Signer>` in a fixed order (delegated, Ed25519, WebAuthn). The order is documented in the generated module docs whenever it matters.
- Authorization is either n-of-n (no policy), a simple threshold (m-of-n multisig), or a weighted threshold whose `weights` argument is positionally aligned with the assembled signer order. A length mismatch panics with a dedicated contract error.
- Verifier addresses, policy addresses and the threshold are deployment arguments rather than baked-in constants, so a single build works on any network.
- An execution entry point option lets the account call contracts it owns, such as its own policies, and an upgradeable option is authorized by the account itself rather than by an access control role.
- The Account type has no access control or explicit trait implementation options, since every `SmartAccount` method authorizes through the account's own context rules.

## Integration

Wired into every consumer of the core package: CLI registry, MCP tool (including the interactive MCP App added in #828), AI assistant function definitions and schemas, and the Stellar UI tab.

## Testing

- Snapshot and API-equivalence tests over the option matrix, plus validation tests for the "at least one signer type" rule.
- Compile tests build and `cargo test` the generated zip for six option combinations. The zip's `test.rs` template generates mock verifier and policy contracts, which the account needs at construction time: `add_context_rule` installs every policy and canonicalizes every external signer key through its verifier.

## Notes for reviewers

- Shared printer changes (`print.ts`, `contract.ts`) were needed for multi-line constructor bodies, commented constructor arguments and multiple associated types. No other kind's snapshots changed.
- The generated code and UI point at the Stellar Registry as a place to look for verifier and policy contracts. That registry is a third party; the choice of which contract to trust is left to whoever deploys.


packages/core/stellar/src/generate/account.ts
```diff
@@ -0,0 +1,21 @@
...
+  webauthnSigners: booleans,
+  policy: policyOptions,
+  executionEntryPoint: booleans,
+  upgradeable: booleans,
+  info: infoOptions,
+};
+
+export function* generateAccountOptions(): Generator<Required<AccountOptions>> {
+  yield* generateAlternatives(blueprint);
+}
```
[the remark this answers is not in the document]
brozorec: Not changing this — the generator behaving this way is the deliberate repo-wide pattern, not an oversight.

**There are no other consumers to protect.** `generateAccountOptions` is not exported from `packages/core/stellar/src/index.ts`, so it is internal to the package. Its only caller is `generateOptions` in `generate/sources.ts`, and `generateContractSubset` there catches `OptionsError` and `continue`s. The premise that "other consumers must handle it" does not hold today, and the return type `Generator<Required<AccountOptions>>` promises completeness of the option set, not validity of every combination.

**Enumerate-then-validate is the architecture in all six language packages.** `generate/sources.ts` has the identical catch in `solidity`, `cairo`, `cairo_alpha`, `stylus`, `confidential` and `stellar`. Blueprints deliberately emit the full cartesian product and the builder is the single source of truth for what is valid.

**Account relies on it far less than existing kinds do.** Counting rejected combinations per Stellar kind:

| Kind | Combinations | Rejected by `OptionsError` |
| --- | --- | --- |
| NonFungible | 3072 | 1824 (59%) |
| Stablecoin | 2304 | 768 (33%) |
| **Account** | **192** | **24 (12.5%)** |
| Fungible / Vault / Governor | 768 / 96 / 48 | 0 |

Pre-filtering only the Account blueprint would make it the single kind that pre-filters, would not achieve consistency (NonFungible and Stablecoin depend on the catch far more heavily), and would restate `buildAccount`'s "at least one signer type" rule in a second place where the two can silently drift. The rule is already covered by two unit tests in `account.test.ts`.

packages/ui/api/ai-assistant/types/languages.ts
```diff
@@ -42,7 +42,10 @@ export type LanguagesContractsOptions = {
   cairoAlpha: CairoAlphaKindedOptions;
   confidential: ConfidentialKindedOptions;
   polkadot: Omit<SolidityContractsOptions, 'Account'>;
-  stellar: Omit<StellarKindedOptions, 'Fungible' | 'NonFungible' | 'Stablecoin' | 'Vault' | 'Governor'> & {
+  stellar: Omit<StellarKindedOptions, 'Account' | 'Fungible' | 'NonFungible' | 'Stablecoin' | 'Vault' | 'Governor'> & {
+    // A smart account has no `access` option, so it cannot take
+    // `StellarCommonContractOptions` wholesale.
+    Account: StellarKindedOptions['Account'] & { upgradeable?: false };
```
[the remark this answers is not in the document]
brozorec: Not changing this either — both halves of the finding do not hold up.

**There is no type conflict with the `boolean` property.** `AiFunctionType<TType>` declares `type?: StringifyPrimaryType<TType>`, and `StringifyPrimaryType` maps via `TType extends boolean ? 'boolean'`. Since `false extends boolean` is true, `StringifyPrimaryType<false>` resolves to `'boolean'`, so declaring `upgradeable: { type: 'boolean' }` against a property typed `false` is valid. That is why `deno-check` passes on this PR; the two files are not in disagreement.

**It does not prevent AI-assisted upgradeable Accounts.** These types are compile-time only. `AiFunctionCall.arguments` is a type annotation over the result of `JSON.parse` on the model's output — there is no zod or other runtime validation of function-call arguments anywhere in `packages/ui/api`. What the model is actually allowed to send is governed by the function definition, which does advertise `upgradeable` as a boolean, and a returned `true` flows through to `buildAccount`, which supports it.

**The restriction matches every sibling.** All five other Stellar kinds get `upgradeable?: false` through `StellarCommonContractOptions`, and `solidity` `Stablecoin` and `Account` already carry the exact same pairing on `master` — restricted in `languages.ts` while their function definitions advertise a boolean. Relaxing it here would make Stellar `Account` the only kind that deviates, with no functional change.

If the `upgradeable?: false` convention is in fact stale across Stellar and Solidity, that is worth a dedicated cleanup PR covering all of them rather than a one-kind exception here.

packages/common/src/ai/descriptions/stellar.ts
```diff
@@ -2,6 +2,8 @@
 // which have different requirements for file extensions in import statements.
 
 export const stellarPrompts = {
+  Account:
+    'Make a smart account (smart wallet) that authorizes operations through configurable signers and policies, for example a multisig with passkey, Ed25519 and delegated signers.',
```
ericglau: We should avoid the term 'wallet' when referring to smart accounts to avoid confusion with the software for interacting with them.
```suggestion
    'Make a smart account contract that authorizes operations through configurable signers and policies, for example a multisig with passkey, Ed25519 and delegated signers.',
```

packages/core/stellar/src/account.ts
```diff
@@ -0,0 +1,340 @@
...
+  if (signerLabels.length > 1 || opts.policy === 'weighted-threshold') {
+    c.addDocumentation('');
+    c.addDocumentation(
+      `Signers are assembled from the constructor arguments in this order: ${signerLabels.join(', ')}.`,
+    );
+  }
+
+  if (opts.policy === 'weighted-threshold') {
+    c.addDocumentation('`weights` is positionally aligned with that signer order.');
+  }
```
ericglau: nit: Is it common practice to document constructor arguments at the contract level doc comments, rather than at the constructor itself?

packages/core/stellar/src/zip-shared.ts
```diff
@@ -101,9 +101,237 @@ fn initial_state() {
...
+  // A WebAuthn key is the 65-byte secp256r1 public key followed by an optional
+  // credential id; canonicalization reads the first 65 bytes.
+  webauthn_keys: () => 'Vec::from_array(&env, [Bytes::from_slice(&env, &[2u8; 65])])',
+  weights: signerCount => `Vec::from_array(&env, [${Array<string>(signerCount).fill('1u32').join(', ')}])`,
+  threshold: () => '1u32',
+  threshold_policy: () => 'policy',
+};
+
+const MOCK_ED25519_VERIFIER = `
+// Minimal Ed25519 verifier contract.
```
ericglau: Are these mock contracts meant to be usable/adaptable for production purposes as well?  If not, we should clearly document that they are for testing purposes only.  Similar for the WebAuthn and Policy ones below.

.changeset/add-stellar-account.md
```diff
@@ -0,0 +1,13 @@
...
+- Each enabled signer type contributes its own typed constructor arguments, which the constructor assembles into a single `Vec<Signer>` in a fixed order (delegated, Ed25519, WebAuthn).
+- Authorization is either n-of-n (no policy), a simple threshold (m-of-n multisig), or a weighted threshold whose `weights` argument is positionally aligned with the assembled signer order.
+- Verifier addresses, policy addresses and the threshold are deployment arguments rather than baked-in constants, so a single build works on any network. The generated code points at the Stellar Registry (https://testnet.rgstry.xyz and https://stellar.rgstry.xyz) as a place to look for verifier and policy contracts, leaving the choice of which to trust to whoever deploys.
+- Adds an execution entry point option, which lets the account call contracts it owns such as its own policies, and an upgradeable option authorized by the account itself.
+- The Account type has no access control or explicit trait implementation options, since every `SmartAccount` method authorizes through the account's own context rules.
```
ericglau: nit: Changelog entries should just briefly outline the main changes.  Feature-level documentation seems too excessive and should belong in the PR description instead (the PR link will be added to the first line during publish, so the user can click through if they are interested).

ericglau (approved):
LGTM overall in terms of the Wizard integration aspects.  Just a few minor comments, thanks.

packages/core/stellar/src/account.ts
brozorec: good point

PR #839, opened 2026-08-06
Feat/stellar :  added validation policies for Governor voting parameters

mesayanroy:
## 📌 Overview
This pull request introduces strict, field-level configuration validation for key governance parameters (Voting Delay, Voting Period, and Proposal Threshold) within @openzeppelin/wizard-stellar.

##  What Was The Issue?
Prior to this change, the Stellar Contracts Wizard UI allowed users to enter invalid governance configurations—such as setting a votingDelay of 0 or setting a votingPeriod smaller than or equal to votingDelay—without displaying inline validation errors.

Generating a Soroban smart contract with invalid governance configurations presents severe issues:

1) Security & Protocol Risks: A Governor deployed with zero voting delay allows instant voting execution without giving token holders or delegates notice to review proposals.
2) Governance Deadlocks: Setting a votingPeriod less than or equal to votingDelay creates an unexecutable contract where voting closes before or at the same time it opens.
3) Suboptimal Developer Experience: Developers were only alerted to invalid governance parameters during downstream compilation or mainnet/testnet deployment.

## What Is The Solution?
We implemented comprehensive, field-scoped validation logic inside buildGovernor within @openzeppelin/wizard-stellar, backed by unit tests:

- Voting Delay: Enforced a minimum requirement of $\ge 1$ ledger (votingDelay >= 1n). If $< 1$, attaches the error message 'Voting delay must be at least 1 ledger' to errors.votingDelay.
- Voting Period: Enforced that Voting Period must be strictly greater than Voting Delay ($votingPeriod > votingDelay$). If $votingPeriod \le votingDelay$, attaches the error message 'Voting period must be greater than voting delay' to errors.votingPeriod.
- Proposal Threshold: Enforced non-negative u128 integer parsing.
- Unified Error Feedback: Accumulated field validation errors into an OptionsError instance, populating inline error tooltips (use:error) in GovernorControls.svelte and automatically disabling export/download buttons until configuration errors are resolved.

## 🛠️ How We Solved It
1. Core Generator Engine (packages/core/stellar/src/governor.ts)
Updated buildGovernor(opts) to validate values and collect field-mapped error messages:

const errors: OptionsErrorMessages = {};

let votingDelay = 0n;
try {
  votingDelay = toUint(allOpts.votingDelay, 'votingDelay', 'u32');
  if (votingDelay < 1n) {
    errors.votingDelay = 'Voting delay must be at least 1 ledger';
  }
} catch (e: unknown) {
  if (e instanceof OptionsError) Object.assign(errors, e.messages);
  else throw e;
}

let votingPeriod = 0n;
try {
  votingPeriod = toUint(allOpts.votingPeriod, 'votingPeriod', 'u32');
  if (votingDelay >= 1n && votingPeriod <= votingDelay) {
    errors.votingPeriod = 'Voting period must be greater than voting delay';
  }
} catch (e: unknown) {
  if (e instanceof OptionsError) Object.assign(errors, e.messages);
  else throw e;
}

if (Object.keys(errors).length > 0) {
  throw new OptionsError(errors);
}

2. Unit Testing Suite (packages/core/stellar/src/governor.test.ts)
Added AVA unit tests verifying that invalid configurations throw appropriate error payloads:

- governor validation votingDelay < 1
- governor validation votingPeriod <= votingDelay
- governor validation invalid proposalThreshold
>All 13 unit tests pass cleanly:

## Why Is This So Important?
- Prevents On-Chain Security Flaws: Governance smart contracts control protocol upgrades and DAOs. Catching invalid parameters at the wizard generator level prevents broken contracts from being deployed on the Stellar network.
- Enhances Developer Productivity: Instant visual feedback via red error tooltips guides developers to construct valid governance parameters without reading raw error logs.
- Upholds OpenZeppelin Quality Standards: Ensures the Stellar Soroban Wizard adheres to the same rigorous validation standards established across OpenZeppelin's Solidity and Cairo wizards
image

### DIFF
--- a/packages/core/stellar/src/governor.test.ts
+++ b/packages/core/stellar/src/governor.test.ts
@@ -81,3 +81,30 @@ testAPIEquivalence('governor API custom', {
 test('governor API assert defaults', async t => {
   t.is(governor.print(governor.defaults), governor.print());
 });
+
+test('governor validation votingDelay < 1', t => {
+  const err = t.throws(() => {
+    buildGovernor({ name: 'MyGovernor', votingDelay: '0' });
+  });
+  t.is(err?.messages?.votingDelay, 'Voting delay must be at least 1 ledger');
+});
+
+test('governor validation votingPeriod <= votingDelay', t => {
+  const err = t.throws(() => {
+    buildGovernor({ name: 'MyGovernor', votingDelay: '100', votingPeriod: '50' });
+  });
+  t.is(err?.messages?.votingPeriod, 'Voting period must be greater than voting delay');
+
+  const errEqual = t.throws(() => {
+    buildGovernor({ name: 'MyGovernor', votingDelay: '100', votingPeriod: '100' });
+  });
+  t.is(errEqual?.messages?.votingPeriod, 'Voting period must be greater than voting delay');
+});
+
+test('governor validation invalid proposalThreshold', t => {
+  const err = t.throws(() => {
+    buildGovernor({ name: 'MyGovernor', proposalThreshold: '-10' });
+  });
+  t.is(err?.messages?.proposalThreshold, 'Not a valid number');
+});
+
--- a/packages/core/stellar/src/governor.ts
+++ b/packages/core/stellar/src/governor.ts
@@ -5,6 +5,8 @@ import { printContract } from './print';
 import { setAccessControl } from './set-access-control';
 import { addUpgradeable } from './add-upgradeable';
 import { setInfo } from './set-info';
+import type { OptionsErrorMessages } from './error';
+import { OptionsError } from './error';
 import { toByteArray, toUint } from './utils/convert-strings';
 import { defineFunctions } from './utils/define-functions';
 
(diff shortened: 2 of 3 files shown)


ozgunozerk:
Thanks for the detailed write-up, @mesayanroy. There's a real issue buried in here, though I think it's a different one than the validation rules target. Let me walk through it.

**The schedule is sequential, not parallel**

In the Stellar governor the two values are laid end to end rather than measured from the same starting point (`packages/governance/src/governor/storage.rs` in `stellar-contracts`):

```rust
vote_snapshot = current_ledger + voting_delay
vote_end      = vote_snapshot  + voting_period
```

`vote_end` builds on `vote_snapshot`, so the voting window always gets its full `voting_period` length no matter what the delay is:

```
created           voting opens        voting closes
ledger 1000       ledger 1011         ledger 1110
   |-----------------|--------------------|
   |   delay = 10    |    period = 100    |
```

**On `votingPeriod > votingDelay`**

Because of that, `votingPeriod <= votingDelay` doesn't close the vote early. `delay = 100, period = 5` puts the snapshot at ledger 1100 and the end at 1105, so voting runs from 1101 to 1105. That's a short vote after a long notice period, which is a configuration some DAOs deliberately want. I ran it against the library and the proposal goes Active, accepts votes, and reaches `Succeeded` as expected. So this rule would reject valid setups.

**On `votingDelay >= 1`**

The first state check is `if current_ledger <= vote_snapshot { Pending }`, using `<=` rather than `<`. With `delay = 0` the snapshot lands on the creation ledger, the proposal stays Pending for that ledger, and voting opens on the next one with the full period intact. Solidity's Governor and the Solidity wizard both allow a zero delay too, so enforcing a minimum here would leave the two wizards disagreeing about what a valid governor looks like. The underlying concern (no notice window) is a genuine tradeoff, but it's the deployer's call to make.

**What you did find: `votingPeriod = 0`**

This one is a real bug, and I'm glad the PR surfaced it. With `period = 0`, `vote_end == vote_snapshot`, and the state machine checks Pending first:

```rust
if current_ledger <= vote_snapshot { return Pending; }
if current_ledger <= vote_end      { return Active;  }
```

Reaching Active would need `current > vote_snapshot` and `current <= vote_end` at the same time, impossible when the two are equal. The proposal is never Active for even a single ledger, so no vote can ever be cast and it drops straight to Defeated. Every proposal, permanently. And since the generated contract sets `voting_period` only in the constructor and exposes no setter, there's no way to recover short of redeploying.

Solidity heads this off at the contract level: `GovernorSettings._setVotingPeriod` reverts with `GovernorInvalidVotingPeriod(0)`, so a bad deploy just fails. That's why the Solidity wizard doesn't carry the check itself.

**Where we'd like the fix to live**

Since this is a contract-level gap, we'd rather close it in `stellar-contracts` so it covers everyone using the library directly and not only wizard users. I've opened OpenZeppelin/stellar-contracts#834 for that, crediting this PR. A wizard-side `votingPeriod >= 1` check is still worth having for the inline feedback, and we may add it alongside.

One small note: the `proposalThreshold` change is already handled by the existing `toUint(..., 'u128')` call, which rejects negatives and non-integers.

So we're going to close this one, but the `votingPeriod = 0` catch is a good find. Thanks for taking the time on it.

